Issue #8, opened 2021-06-03
Client: implement retry logic for TO1


puiterwijk:
At this moment, the client attempts TO1 once, with the very first client URL retrieved from the rendezvous info.
It should perform the constant retries on the different entries, taking the delays into account.

PR #35, opened 2021-07-15
Client - Split RV & TO1. Support multiple rendezvous servers

jnogol:
Related to #8, will make easier implementing the retries.

Changes:
- Support for multiple rendezvous servers in the config file, and rename `get_client` to `get_client_list`.
- Split `get_to1d_from_rv` into `get_rv_info` and `get_to1d`.
- In the testing part:
  - Add `-e` flag in `integration/test.sh`.
  - Add one more rendezvous server (that won't be operating, just for the sake of dealing with multiple RV servers) in `integration/testdata/rendezvous-info.yml`.

### DIFF
--- a/client-linuxapp/src/main.rs
+++ b/client-linuxapp/src/main.rs
@@ -60,7 +60,13 @@ fn get_to2_urls(entries: &[TO2AddressEntry]) -> Vec<String> {
     urls
 }
 
-async fn get_client(rv_info: Vec<RendezvousInterpretedDirective>) -> Result<ServiceClient> {
+async fn get_client_list(
+    rv_info: Vec<RendezvousInterpretedDirective>,
+) -> Result<Vec<ServiceClient>> {
+    if rv_info.is_empty() {
+        bail!("No rendezvous entries found we can construct a client for");
+    }
+    let mut service_client_list = Vec::new();
     for rv_entry in rv_info {
         let urls = rv_entry.get_urls();
         if urls.is_empty() {
@@ -75,11 +81,12 @@ async fn get_client(rv_info: Vec<RendezvousInterpretedDirective>) -> Result<Serv
         if rv_entry.user_input {
             todo!();
         }
-        return Ok(fdo_http_wrapper::client::ServiceClient::new(
-            &urls.first().unwrap(),
-        ));
+        for url in &urls {
+            service_client_list.push(fdo_http_wrapper::client::ServiceClient::new(&url));
+        }
     }
-    bail!("No rendezvous entries found we can construct a client for");
+
+    Ok(service_client_list)
 }
 
 async fn perform_to1(
@@ -135,7 +142,7 @@ async fn perform_to1(
     Ok(rv_redirect.into_to1d())
 }
 
-async fn get_to1d_from_rv(devcred: &dyn DeviceCredential) -> Result<COSESign> {
+fn get_rv_info(devcred: &dyn DeviceCredential) -> Result<Vec<RendezvousInterpretedDirective>> {
     // Determine RV info
     let rv_info = devcred
         .rendezvous_info()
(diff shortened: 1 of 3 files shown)


client-linuxapp/src/main.rs
```diff
@@ -75,11 +81,15 @@ async fn get_client(rv_info: Vec<RendezvousInterpretedDirective>) -> Result<Serv
...
-            &urls.first().unwrap(),
-        ));
+        for url in &urls {
+            service_client_list.push(fdo_http_wrapper::client::ServiceClient::new(&url));
+        }
     }
-    bail!("No rendezvous entries found we can construct a client for");
+
+    // THIS IS JUST FOR TEST, REMOVE LATER
+    service_client_list.reverse();
```
jnogol, line 90: This is just for reviewers to see what would happen if the client encounters a non-reachable rendezvous server. Once the review is approved, I'll get rid of this.

puiterwijk: Could you get rid of this, since otherwise the automation will merge this in?
You could just swap around the two RV entries to demonstrate this.

jnogol: Swapping the entries doesn't work, seems to be a bug probably caused by https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/owner-tool/src/main.rs#L813. In any case, just deleted it and pushed the change.

client-linuxapp/src/main.rs
```diff
@@ -425,10 +445,15 @@ async fn main() -> Result<()> {
...
+    // Get rv entries
+    let rv_info = get_rv_info(dc.as_ref());
+
+    let client_list = get_client_list(rv_info.unwrap())
         .await
-        .context("Error getting to1d from rendezvous server")?;
+        .context("Error getting usable rendezvous client")?;
+
+    // Get owner info
+    let to1d = get_to1d(dc.as_ref(), client_list).await.unwrap();
```
puiterwijk, line 456: Calling `.unwrap()` is a bad idea.

client-linuxapp/src/main.rs
```diff
@@ -145,15 +152,28 @@ async fn get_to1d_from_rv(devcred: &dyn DeviceCredential) -> Result<COSESign> {
...
+            Ok(to1) => {
+                return Ok(to1);
+            }
+            Err(e) => {
+                log::trace!("{} with {:?}", e, client);
+                continue;
+            }
+        }
+    }
+    bail!("no client replied!")
```
puiterwijk, line 176: Was this supposed to say `No server replied`?

client-linuxapp/src/main.rs
```diff
@@ -425,10 +445,15 @@ async fn main() -> Result<()> {
         return Ok(());
     }
 
-    // Get owner info
-    let to1d = get_to1d_from_rv(dc.as_ref())
+    // Get rv entries
+    let rv_info = get_rv_info(dc.as_ref());
+
+    let client_list = get_client_list(rv_info.unwrap())
```
puiterwijk, line 451: Getting a full client list before going over it might make the processing of other RV entries a bit more complicated, since it detaches the looping over RV entries from the building of clients.

puiterwijk (changes requested):

client-linuxapp/src/main.rs
jnogol: The reason of doing this was that I would get a `Result<COSESign, Error>` instead of a `COSESign`. What should I do instead of using `.unwrap()` and why is it a bad idea?

runcom: http://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/error-handling.html#unwrapping-explained

Basically it's bad to just panic and exit here - you can `match` instead and handle any error

client-linuxapp/src/main.rs
jnogol: `Couldn't get TO1 from any Rendezvous server?` maybe? Bear in mind that I want to implement retries after this PR.

client-linuxapp/src/main.rs
jnogol: What can we do as an alternative?

client-linuxapp/src/main.rs
puiterwijk: Or you can use [the `?` operator](https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html).

client-linuxapp/src/main.rs
puiterwijk: Right. I just figured the exactly message might be a bit off, but this is just a nitpick.

client-linuxapp/src/main.rs
jnogol: `?` worked like a charm. Thank you both!

client-linuxapp/src/main.rs
jnogol: In the next PR, which I plan to be focused exclusively on the retries, what I can do is loop over RV entries and, instead of passing a vector of `RendezvousInterpretedDirective` to `get_client_list`, just pass one entry to it (so `RendezvousInterpretedDirective` instead of `Vec<RendezvousInterpretedDirective>`.

What do you think? @puiterwijk

client-linuxapp/src/main.rs
jnogol: I think it can be more descriptive too. WDYT about `Couldn't get TO1 from any Rendezvous server?` @puiterwijk

client-linuxapp/src/main.rs
puiterwijk: That sounds decent.

puiterwijk (approved):

PR #38, opened 2021-07-27
Dump JSON

Avielyo10:
The current dump is pseudo YAML which makes it hard to parse.


puiterwijk:
I would very much prefer this to be a flag, or to use YAML: the `dump-` commands are primarily intended for manual use. And while I'm not opposed to making it machine-parseable, I don't want to lose the ability to read it by hand.

puiterwijk:
Additionally, if you change one, please also change the other `dump-` command.

Avielyo10:
@puiterwijk Ok, I'll

PR #39, opened 2021-07-28
Settings change

Avielyo10:
- Moving configuration file to be optional will make it easier on deployments, also will make it easier by the user to catch the
configuration he's missing.
- Change prefix to fit bash/zsh export.


puiterwijk (approved):
LGTM

Issue #29, opened 2021-07-06
Client: TO2 retries


puiterwijk:
At this moment, the client only ever uses the first URL returned in the TO1d data.
It should retry this with the different URLs retrieved from TO1d.

jnogol:
Fixed via #41

PR #41, opened 2021-08-18
Implement TO1 and TO2 retries logic #8 #29

### LINKED ISSUE #8: Client: implement retry logic for TO1

jnogol:
Signed-off-by: Jose Noguera <jnoguera@redhat.com>

Closes #8
Related to #29

### DIFF
--- a/client-linuxapp/src/main.rs
+++ b/client-linuxapp/src/main.rs
@@ -60,32 +60,25 @@ fn get_to2_urls(entries: &[TO2AddressEntry]) -> Vec<String> {
     urls
 }
 
-async fn get_client_list(
-    rv_info: Vec<RendezvousInterpretedDirective>,
-) -> Result<Vec<ServiceClient>> {
-    if rv_info.is_empty() {
-        bail!("No rendezvous entries found we can construct a client for");
-    }
+async fn get_client_list(rv_entry: &RendezvousInterpretedDirective) -> Result<Vec<ServiceClient>> {
     let mut service_client_list = Vec::new();
-    for rv_entry in rv_info {
-        let urls = rv_entry.get_urls();
-        if urls.is_empty() {
-            continue;
-        }
-        if rv_entry.bypass {
-            todo!();
-        }
-        if rv_entry.wifi_ssid.is_some() {
-            todo!();
-        }
-        if rv_entry.user_input {
-            todo!();
-        }
-        for url in &urls {
-            service_client_list.push(fdo_http_wrapper::client::ServiceClient::new(url));
-        }
-    }
 
+    let urls = rv_entry.get_urls();
+    if urls.is_empty() {
+        bail!("No URLs found in this RV_entry");
+    }
+    if rv_entry.bypass {
+        todo!();
+    }
+    if rv_entry.wifi_ssid.is_some() {
+        todo!();
+    }
+    if rv_entry.user_input {
+        todo!();
+    }
+    for url in &urls {
+        service_client_list.push(fdo_http_wrapper::client::ServiceClient::new(url));
+    }
     Ok(service_client_list)
 }
 
@@ -127,7 +120,7 @@ async fn perform_to1(
             .context("Error creating EATokenPayload")?;
 
     // Create signature over nonce4
-    let signer = devcred.get_signer().context("Error getting COSE signer")?;
+    let signer = devcred.get_signer().context("Error getting Cose signer")?;
     let token =
         COSESign::from_eat(eat, None, signer.as_ref()).context("Error signing new token")?;
     log::trace!("Sending token: {:?}", token);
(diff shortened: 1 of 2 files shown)


client-linuxapp/src/main.rs
```diff
@@ -446,36 +439,55 @@ async fn main() -> Result<()> {
...
+    let mut rv_entry_it = rv_info.iter().cycle();
 
-    if to2_addresses.is_empty() {
-        bail!("No valid TO2 addresses received");
-    }
-
-    for to2_address in to2_addresses {
-        match perform_to2(devcred_location.borrow(), dc.as_ref(), &to2_address)
+    loop {
+        let rv_entry = rv_entry_it.next().unwrap();
```
jnogol, line 446: @puiterwijk I know you're not a big fan of `unwrap`, but tried with `?` here, didn't work, and couldn't find other way to get this done.

client-linuxapp/src/main.rs
```diff
@@ -446,37 +484,62 @@ async fn main() -> Result<()> {
...
 
-    let to1d_payload: UnverifiedValue<TO1DataPayload> = to1d
-        .get_payload_unverified()
-        .context("Error getting the TO2 payload")?;
-    let to2_addresses = to1d_payload.get_unverified_value().to2_addresses();
-    let to2_addresses = get_to2_urls(to2_addresses);
-    log::info!("Got TO2 addresses: {:?}", to2_addresses);
+        let client_list = get_client_list(rv_entry)
+            .await
+            .context("Error getting usable rendezvous client")?;
```
puiterwijk, line 498: This will return an `Error` if an RV Entry only specifies wifi or delay, and no URL.

client-linuxapp/src/main.rs
```diff
@@ -60,32 +67,25 @@ fn get_to2_urls(entries: &[TO2AddressEntry]) -> Vec<String> {
...
-            todo!();
-        }
-        for url in &urls {
-            service_client_list.push(fdo_http_wrapper::client::ServiceClient::new(url));
-        }
-    }
 
+    let urls = rv_entry.get_urls();
+    if urls.is_empty() {
+        bail!("No URLs found in this RV_entry");
```
puiterwijk, line 75: This means that RV Entries that only have wifi info or request user input are not processed.

client-linuxapp/src/main.rs
```diff
@@ -446,37 +484,62 @@ async fn main() -> Result<()> {
...
+        let to1d = get_to1d(dc.as_ref(), client_list).await;
+        if let Err(e) = to1d {
+            log::trace!("Error getting TO1d {:?} with rv_entry {:?}", e, rv_entry);
+            rv_entries_processed =
+                sleep_if_needed(rv_entries_processed, rv_info.len(), rv_entry.delay);
+            continue;
+        }
+        let to1d_payload: UnverifiedValue<TO1DataPayload> = to1d?
+            .get_payload_unverified()
+            .context("Error getting the TO1 payload")?;
```
puiterwijk, line 511: On errors we need to retry, not cancel.

client-linuxapp/src/main.rs
```diff
@@ -446,37 +484,62 @@ async fn main() -> Result<()> {
...
-        match perform_to2(devcred_location.borrow(), dc.as_ref(), &to2_address)
-            .await
-            .context("Error performing TO2 ownership protocol")
-        {
-            Ok(_) => break,
-            Err(e) => {
-                log::trace!("{:?} with TO2 address {}", e, to2_address);
-                continue;
+        if to2_addresses.is_empty() {
+            bail!("No valid TO2 addresses received");
```
puiterwijk, line 519: In this case, we need to continue to the next RV_info entry.

puiterwijk (changes requested):
I think that rather than the complicated dance with `rv_entries_processed` and the cycling iterator, maybe it's more readable to instead go for a structure where you have e.g.:
```
loop {
    for rv_entry in rv_info_it.clone() {
    }
    // Perform sleep here
}
```

client-linuxapp/src/main.rs
jnogol: Tru dat. instead of bailing, maybe `continue`?

client-linuxapp/src/main.rs
jnogol: Ok, so instead of bailing, logging the error + `continue`?

client-linuxapp/src/main.rs
jnogol: I understand that with [this](https://github.com/fedora-iot/fido-device-onboard-rs/pull/41/files#r693861156), it'd be addressed, am I right?

puiterwijk (approved):
LGTM, thanks!

client-linuxapp/src/main.rs
puiterwijk: Yep, thanks.

PR #44, opened 2021-08-25
Implement open modes for store

puiterwijk:
This allows having a store that's marked read-only or write-only and
have this enforced by the type system.


puiterwijk:
This is specifically for the manufacturing server, which will have one read-only store (mfgstring -> public key) and one write-only store (new ownership vouchers).

runcom (approved):

PR #45, opened 2021-08-26
Owner Tool: Run TO0 to every RV server listed in the RV info

### LINKED ISSUE #36: Owner Tool: Run TO0 to every RV server listed in the RV info
At the moment, the owner tool only grabs the first rendezvous server that it finds in the voucher: https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/owner-tool/src/main.rs#L813-L818

jnogol:
Closes #36

Changes:
- Reformatted test script and added delay and inverted order in rendezvous test info file
- Working TO0 with multiple entries

### DIFF
--- a/owner-tool/src/main.rs
+++ b/owner-tool/src/main.rs
@@ -797,55 +797,72 @@ async fn report_to_rendezvous(matches: &ArgMatches<'_>) -> Result<(), Error> {
     let owner_addresses: Result<Vec<Vec<TO2AddressEntry>>> =
         owner_addresses.drain(..).map(|v| v.try_into()).collect();
     let mut owner_addresses = owner_addresses.context("Error parsing owner addresses")?;
-    let owner_addresses = owner_addresses.drain(..).flatten().collect();
+    let owner_addresses: Vec<TO2AddressEntry> = owner_addresses.drain(..).flatten().collect();
 
     // Determine the RV IP
-    let rv_info = ov
-        .get_header()
-        .context("Error getting OV header")?
+    let rv_info = ov_header
         .rendezvous_info
         .to_interpreted(RendezvousInterpreterSide::Owner)
         .context("Error parsing rendezvous directives")?;
     if rv_info.is_empty() {
         bail!("No rendezvous information found that's usable for the owner");
     }
-    // Use the first entry
-    let rv_info = rv_info.first().unwrap();
-    let rv_urls = rv_info.get_urls();
-    if rv_urls.is_empty() {
-        bail!("No usable rendezvous URLs were found");
+    let mut rendezvous_performed = false;
+    for rv_directive in rv_info {
+        let rv_urls = rv_directive.get_urls();
+        if rv_urls.is_empty() {
+            log::info!(
+                "No usable rendezvous URLs were found for RV directive: {:?}",
+                rv_directive
+            );
+            continue;
+        }
+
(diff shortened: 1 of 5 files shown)


owner-tool/src/main.rs
```diff
@@ -797,55 +797,69 @@ async fn report_to_rendezvous(matches: &ArgMatches<'_>) -> Result<(), Error> {
...
-        bail!("No usable rendezvous URLs were found");
-    }
-    let rv_url = rv_urls.first().unwrap();
-
-    println!("Using rendezvous server at url {}", rv_url);
+    let mut rendezvous_performed = false;
+    for rv_directive in rv_info {
+        let rv_urls = rv_directive.get_urls();
+        if rv_urls.is_empty() {
+            bail!("No usable rendezvous URLs were found");
```
puiterwijk, line 814: I think that we might want to just continue if we don’t get useful URLs. It could be that that specific entry only contained device-specific items.

owner-tool/src/main.rs
```diff
@@ -797,55 +797,69 @@ async fn report_to_rendezvous(matches: &ArgMatches<'_>) -> Result<(), Error> {
...
-    // Send: OwnerSign, Receive: AcceptOwner
-    let accept_owner: RequestResult<messages::to0::AcceptOwner> = rv_client
-        .send_request(messages::to0::OwnerSign::new(to0d, to1d), None)
-        .await;
-    let accept_owner = accept_owner.context("Error registering self to rendezvous server")?;
+            let hello_ack = match hello_ack.context("Error requesting nonce from rendezvous server")
+            {
+                Ok(hello_ack) => hello_ack,
+                Err(e) => {
+                    println!("We're inside the error: {:?}", e);
```
puiterwijk, line 831: Let’s use `log::info!` for this, and provide a but more information on what we were trying to accomplish?

puiterwijk (changes requested):
Looks decent overall, just a couple of suggestions.

owner-tool/src/main.rs
jnogol: Oh, that was me forgetting a debugging message lol.

In order to use `log::info`, we'll need to add `log` and `pretty_env_logger` as dependencies. If it's OK, what I can do is create a different PR and migrate all the `println` in the Owner Tool to `log::{info, error, trace}`, etc.

In any case, what do you think of the following match statement?

```
let hello_ack = match hello_ack
            {
                Ok(hello_ack) => hello_ack,
                Err(e) => {
                    log::error!("Error requesting nonce from rendezvous server: {:?}", e);
                    continue;
                }
            };
```

owner-tool/src/main.rs
jnogol: Log an error/trace and continue to the next entry? Does that sound good?

puiterwijk: An `info` I'd say, since it's not an error.

owner-tool/src/main.rs
puiterwijk: Yep, that works. And let's for now just import `log` and not yet `pretty_env_logger`: PR #50 centralizes the initialization of logging.
I don't think we want to replace every `println` with a `log`: the actual output of the command the user executed should stay a `println!`. (but yes, there is a reasonable amount of things that should become `log`).

puiterwijk (approved):
LGTM

PR #46, opened 2021-09-07
Make message ordering and encryption stronger

puiterwijk:
This enforces the ordering of messages within a session, by making both
the client and server check whether the next message is allowed at this
place of the protocol.
This would avoid possible issues where an attacker skips a part of the
process, skipping to the next part of the code.

This patch also includes metadata on the messages whether they're
required to be encrypted, which is checked by client and server.


puiterwijk:
This basically implements an implicit state machine for the client and server, with explicit state transitions.

runcom (approved):
LGTM

PR #48, opened 2021-09-09
Fix PublicKey data structure

puiterwijk:
When doing the initial implementation of PublicKey, we were expecting
X509 certificate DER data to be passed in, but we should instead be
sending and expecting DER encoded public keys.

During this, we also ensured that the DER data is correct at the moment
a PublicKey is constructed.


runcom:
lgtm

runcom (approved):

PR #51, opened 2021-09-21
Implement DI and DIUN

puiterwijk:
This implements the Device Initalize and Device Initialize over
Untrusted Networks (DIUN) protocols.
The DIUN protocol spec is updated to the current version of the draft
protocol, and all the key attestation bits are removed for now.
Key attestation will be brought back in the future, after more
consideration.


puiterwijk:
At this moment, there are no tests yet. I will be making and sending those soon, but this patch set is good for initial review.

puiterwijk:
And the tests are now in place, with a new test framework in Rust! This is now ready for review 😄  (To whomever reviews this: I'm sorry. This turned out to be a massive PR).

Cargo.toml
```diff
@@ -8,4 +8,8 @@ members = [
     "owner-onboarding-service",
     "owner-tool",
     "rendezvous-server",
+    "manufacturing-server",
+    "manufacturing-client",
+
```
jnogol, line 13: just curious, why the empty line?

puiterwijk: Because the integration-tests crate isn't intended to be shipped.

specs/diun.md
```diff
@@ -35,124 +37,144 @@ The Connect message is sent to the Manufacturer to start the key exchange.
...
-Provides a challenge to the Device to prove that the public key is of the claimed type.
-If no challenge is required, the value of `KeyAttestationChallenge` is `null`.
+Provide the key type and key storage types that are acceptable to this manufacturing
+server.
+If no KeyStorageTypes are provided, the client can select any type of key storage
+that it wants.
 
 
-### Step 5: Attestation
+### Step 5: Provide Key, Type 314
```
jnogol, line 142: 314 or 214?

data-formats/src/messages/diun.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+}
+
+impl Message for Connect {
+    fn message_type() -> MessageType {
+        MessageType::DIUNConnect
+    }
+
+    fn is_valid_previous_message(message_type: Option<MessageType>) -> bool {
+        matches!(message_type, None)
```
jnogol, line 57: I understand that what we're ensuring here is that there's no previous message?

puiterwijk: Yep. This is the start of an interaction, so no messages should have been before this during this session.

specs/diun.md
puiterwijk: 214, will fix.

jnogol (approved):
LGTM

PR #56, opened 2021-09-23
Move integration tests from bash to Rust

puiterwijk:
This makes it so `cargo test` runs the integration tests, and it is much simpler to run the tests.
One command to rule them all!


integration-tests/tests/common/mod.rs
```diff
@@ -214,8 +225,16 @@ impl TestContext {
             .map(PathBuf::from)
     }
 
+    pub fn keyspath(&self) -> PathBuf {
```
runcom, line 228: this is super silly not blocking, `keys_path`

integration-tests/tests/common/mod.rs
```diff
@@ -584,22 +666,20 @@ impl TestClientResult {
 #[derive(Debug)]
 pub struct TestServerConfigurator<'a> {
     binary: Binary,
-    test_path: &'a Path,
-    server_path: &'a Path,
+    test_context: &'a TestContext,
+    //server_path: &'a Path,
```
runcom, line 670: drop this?

runcom (approved):

PR #62, opened 2021-09-23
Automatically merge dependabot PRs

puiterwijk:
Automatically merge PRs submitted by dependabot after CI finishes.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -14,3 +14,12 @@ pull_request_rules:
         name: fido-device-onboard-rs
         method: merge
         rebase_fallback: none
+  - name: Automatic merge of dependabot PRs
+    conditions:
+    - base=main
+    - author=dependabot
+    actions:
+      queue:
+        name: fido-device-onboard-rs
+        method: merge
+        rebase_fallback: none


puiterwijk:
This is marked as Draft so people have time to give their opinion on this PR and whether this is a good idea.

runcom:
> This is marked as Draft so people have time to give their opinion on this PR and whether this is a good idea.

I think it's a good idea, not sure about downsides?

puiterwijk:
@runcom theoretically it could introduce a version of a package that we don't actually work with. But I think that that's not a problem, due to:
1. we have (integration) tests
2. if we encounter one of those, what is the chance we would notice it if CI doesn't?

nullr0ute:
I have concerns around "going with the latest" automatically and just "bumping for the sake of it" because that may add a lot of churn for distro packaging in particular, more around Fedora than RHEL. Is there a "only bump if there's a dependency with a CVE" mode because also if we don't bump we may also end up with security vuln. Basically it's a tightrope walk.

puiterwijk:
The problem I would have with that, is that there is no metadata in the crates ecosystem that encodes whether an update contains security fixes.
And even if that were there, an update might later on turn out to fix security issues.
So the alternative is that any time there's a dep update, someone would need to manually check it.

jnogol:
I'm all in for the automerge as there're tests in place, but I can understand @nullr0ute concerns that updating for the sake of updating introduces _some_ uncertainity.

An alternative: as the security-related PRs contain `[Security]` in the PR title and a `security` label, we can expand the conditions to check if the PR has those and then automatically merge.

But again, my vote goes for yes, automerge all Dependabot's PRs.

puiterwijk:
I think that @nullr0ute's concern is valid, and we'll just look at the dependabot PRs as they come in ourselves.
That's also good so that we can decide whether a new release is required due to security things, so I'm going to close this PR.

PR #64, opened 2021-09-29
fix: bring PublicKey back to certificates

puiterwijk:
In PR #48, we changed the PublicKey structure to DER encoded public
keys, because that's how the specification reads.
Unfortunately, that is not the way it was intended: intended was an
actual DER-encoded X509 certificate (X509), or an x5chain (COSEX509).
This will be clarified in an upcoming erratum to the specification.
This patch brings it back to the correct structures.

### DIFF
--- a/data-formats/src/constants.rs
+++ b/data-formats/src/constants.rs
@@ -61,27 +61,30 @@ impl TryFrom<MessageDigest> for HashType {
     }
 }
 
+const RS256: i16 = -257;
+const RS384: i16 = -258;
+
 #[derive(Debug, Clone, Copy, Serialize_repr, Deserialize_repr, Eq, PartialEq)]
-#[repr(i8)]
+#[repr(i16)]
 #[non_exhaustive]
 pub enum DeviceSigType {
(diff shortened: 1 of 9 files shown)


runcom (approved):

runcom:
uhm, now main gives me:

```
➜  fido-device-onboard-rs git:(main) cargo build 
   Compiling fdo-manufacturing-client v0.1.0 (/home/amurdaca/koding/fido-device-onboard-rs/manufacturing-client)
error[E0599]: no method named `raw_private_key` found for struct `PKey<Private>` in the current scope
   --> manufacturing-client/src/main.rs:380:22
    |
380 |                     .raw_private_key()
    |                      ^^^^^^^^^^^^^^^ method not found in `PKey<Private>`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `fdo-manufacturing-client` due to previous error

```

puiterwijk:
@runcom ah. That's unrelated to this specific PR (that would've been the DI PR).
To fix it, remove your `Cargo.lock` and compiled things, and recompile. I'll try to figure out which version of the `openssl` crate added that method, and bump the dependency.

PR #65, opened 2021-09-29
Bump openssl dependency for manufacturing-client

puiterwijk:
We use the openssl PKey::raw_private_key function, as used for hmacs.
That got added in 0.10.36, so actually bump the version dependency.


puiterwijk:
This is replaced by #66, to fix it by just not using that function.

PR #70, opened 2021-10-12
build: add spec to build RPM

runcom:
Fixes https://github.com/fedora-iot/fido-device-onboard-rs/issues/135


fido-device-onboard.spec
```diff
@@ -0,0 +1,65 @@
...
+%install
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-client-linuxapp
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-manufacturing-client
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-manufacturing-server
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-owner-onboarding-server
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-rendezvous-server
+# TODO(runcom): we may want to move this to {_bindir} so admins can use it directly
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-owner-tool
+install -D -m 0644 -t %{buildroot}%{_unitdir} examples/systemd/*
+# TODO(runcom): this has to go under /usr (???)
```
runcom, line 43: cross linking my q from https://github.com/fedora-iot/fido-device-onboard-rs/pull/75/files#r732989096

why `/usr/fdo` - rn I'm putting the configs under `/etc/fdo`

examples/systemd/fdo-owner-onboarding-server.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=FDO owner onboarding server
+After=network-online.target
+
+[Service]
+Environment=LOG_LEVEL=trace
```
puiterwijk, line 6: I don't think we should have `LOG_LEVEL=trace` by default. I think this should be `info`. (We have an outstanding issue to check all log levels).

fido-device-onboard.spec
```diff
@@ -0,0 +1,122 @@
+%define dracutlibdir %{_prefix}/lib/dracut
+%bcond_without check
+%global __cargo_skip_build 0
+%global __cargo_is_lib() false
+%global gitversion 43f7674dad4abc470cdab505fbf49f571539cea8
```
puiterwijk, line 5: Is there any chance that the one we add in the repo has the gitversion automatically determined?

fido-device-onboard.spec
```diff
@@ -0,0 +1,122 @@
...
+
+Name:           fido-device-onboard
+Version:        0.1.0
+Release:        1%{?dist}
+Summary:        An implementation of the FIDO Device Onboard Specification written in rust
+
+License:        BSD 3
+URL:            https://github.com/fedora-iot/fido-device-onboard-rs/
+Source:         %{url}/archive/%{gitversion}/%{name}-rs-%{gitversion}.tar.gz
+Source1:        %{name}-rs-%{gitversion}-vendor.tar.gz
```
puiterwijk, line 15: Given that Fedora doesn't vendor things, maybe make this Source dependent on the distro it's being built for?

fido-device-onboard.spec
```diff
@@ -0,0 +1,122 @@
...
+%prep
+%autosetup -n %{name}-rs-%{gitversion} -p1
+%cargo_prep -V 1
+
+%build
+%cargo_build
+
+%install
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-client-linuxapp
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-manufacturing-client
```
puiterwijk, line 36: Isn't libexecdir only for binaries that are being executed by other binaries?
These are either executed by hand, or by a systemd unit, and I thought that would have to go into /usr/bin?

fido-device-onboard.spec
```diff
@@ -0,0 +1,122 @@
...
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-manufacturing-server
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-owner-onboarding-server
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-rendezvous-server
+# TODO(runcom): we may want to move this to {_bindir} so admins can use it directly
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-owner-tool
```
puiterwijk, line 41: This definitely should go into bindir.

fido-device-onboard.spec
```diff
@@ -0,0 +1,122 @@
...
+install -D -m 0644 -t %{buildroot}%{_unitdir} examples/systemd/*
+install -D -m 0644 -t %{buildroot}%{_datadir}/fdo examples/config/*
```
puiterwijk, line 43: Shouldn't this be `%{_sysconfdir}` (i.e. `/etc`)?

examples/systemd/fdo-owner-onboarding-server.service
runcom: I'll move these to `info`

fido-device-onboard.spec
runcom: not sure I fully understand - this is also just for testing, we're gonna have semver versions here I think

fido-device-onboard.spec
runcom: work to make this spec working on fedora is still ongoing but I don't want to block rhel packaging on that

fido-device-onboard.spec
runcom: not fully sure, that's something I've always done for systemd unit cause those binaries aren't called by users (and systemd isn't a user right?)

fido-device-onboard.spec
runcom: yeep

fido-device-onboard.spec
runcom: we said we wanted these config under `/usr/share` - I'm fine either way, we just have to decide before I change it again :smile:

runcom: also, those are just examples, not even made for actually starting the services (as you need keys and other stuff anyway)

fido-device-onboard.spec
runcom: it's now rhel only but we still need to find a way to add buildrequires for fedora

fido-device-onboard.spec
runcom: ok I understand - how would you do that tho? :thinking:

runcom: we can template the spec and add a Makefile target that does that if we're ever going to support a `make rpm` target - ok I'll do that

runcom: done, there's a `make rpm` now that builds the RPM and populates the commit variable as well

fido-device-onboard.spec
puiterwijk: Ah, right. In that case, maybe they should go under `%{_docdir}`?

nullr0ute: Yes

dracut/52fdo/manufacturing-client.service
```diff
@@ -0,0 +1,17 @@
...
+Before=coreos-installer-reboot.service
+ConditionPathExists=/etc/manufacturing-client-config
+Requires=dev-disk-by\x2dlabel-boot.device
+
+OnFailure=emergency.target
+OnFailureJobMode=replace-irreversibly
+
+[Service]
+Type=oneshot
+EnvironmentFile=/etc/manufacturing-client-config
```
puiterwijk, line 15: Should we add `LOG_LEVEL=info` here as well?
Does systemd just append that to the contents it reads from the EnvironmentFile?

fido-device-onboard.spec
runcom: so `%{_docdir} it is

examples/config/rendezvous-info.yml
```diff
@@ -0,0 +1,5 @@
+---
+- dns: fdo-test.redhat.com
```
puiterwijk, line 2: Let's make this generic? Something like `fdo.example.com`?

examples/config/owner-addresses.yml
```diff
@@ -0,0 +1,5 @@
+---
+- transport: HTTP
+  port: 8081
+  addresses:
+    - dns_name: fdo-test.redhat.com
```
puiterwijk, line 5: Let's make this e.g. `fdo.example.com`?

runcom:
Should really be ready now? @puiterwijk @nullr0ute @jnogol ptal - I want to get this built asap (at least in RHEL) and we can update whatever we want as follow up :)

nullr0ute:
Looking

runcom:
The TL;DR; for Fedora is that we cannot use automatic Requires/BuildRequires as this project uses a workspace and macros aren't supporting that - again, I think this is a great start for packaging and unblocks us from building in RHEL and by using `make rpm` you can also jump on a RHEL machine and build local packages anyway

nullr0ute:
Can we not just smash all of of this into a single commit?

runcom:
> Can we not just smash all of of this into a single commit?

not sure what you mean :thinking: I've split changes into commit as you requested above - something might have slipped but can you clarify?

fido-device-onboard.spec
```diff
@@ -0,0 +1,133 @@
+%define dracutlibdir %{_prefix}/lib/dracut
+%bcond_without check
+%global __cargo_skip_build 0
+%global __cargo_is_lib() false
+%global forgeurl https://github.com/fedora-iot/fido-device-onboard-rs
```
nullr0ute, line 5: It doesn't look like this or the forgemeta below are used anywhere

fido-device-onboard.spec
```diff
@@ -0,0 +1,133 @@
...
+%forgemeta
+
+Name:           fido-device-onboard
+Version:        0.1.0
+Release:        1%{?dist}
+Summary:        An implementation of the FIDO Device Onboard Specification written in rust
+
+License:        BSD 3
+URL:            https://github.com/fedora-iot/fido-device-onboard-rs
+Source:         %{url}/archive/%{commit}/%{name}-rs-%{commit}.tar.gz
```
nullr0ute, line 16: Commit isn't defined anywhere, we should probably just use the version nomenclature everywhere

fido-device-onboard.spec
```diff
@@ -0,0 +1,133 @@
...
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-client-linuxapp
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-manufacturing-client
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-manufacturing-server
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-owner-onboarding-server
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-rendezvous-server
+install -D -m 0755 -t %{buildroot}%{_bindir}/fdo target/release/fdo-owner-tool
+install -D -m 0644 -t %{buildroot}%{_unitdir} examples/systemd/*
+install -D -m 0644 -t %{buildroot}%{_docdir}/fdo examples/config/*
+# 52fdo
+install -D -m 0755 -t %{buildroot}%{dracutlibdir}/modules.d/52fdo dracut/52fdo/module-setup.sh
```
nullr0ute, line 56: I think this should just be called 52fdo IE drop the dracut in the module name as it's implied by the location.

fido-device-onboard.spec
```diff
@@ -0,0 +1,133 @@
...
+
+%files -n fdo-owner-onboarding-server
+%license LICENSE
+%{_libexecdir}/fdo/fdo-owner-onboarding-server
+%{_docdir}/fdo/owner-onboarding-server.yml
+%{_unitdir}/fdo-owner-onboarding-server.service
+
+%package -n fdo-rendezvous-server
+Summary: FDO Rendezvous Server implementation
+License: BSD 3
```
nullr0ute, line 86: Can drop all the License lines from the subpackages, if not there they inherit from the root package. Note the %license lines need to stay

nullr0ute (changes requested):

fido-device-onboard.spec
runcom: they're used to generate metadata as per fedora/rhel guidelines

nullr0ute:
> > Can we not just smash all of of this into a single commit?
> 
> not sure what you mean thinking I've split changes into commit as you requested above - something might have slipped but can you clarify?

Ignore me, I need coffee...

fido-device-onboard.spec
runcom: the Makefile defines it - downstream will add it, this is the very same osbuild/osbuild-composer does so you can just git clone this and build a fully functional RPM

fido-device-onboard.spec
runcom: will do perfect :+1:

runcom:
> Ignore me, I need coffee...

we ALL need coffee :coffee:

fido-device-onboard.spec
runcom: awesome

fido-device-onboard.spec
runcom: there's no `dracut` in the name :thinking:

fido-device-onboard.spec
runcom: done

runcom:
@nullr0ute all comments addressed

Makefile
```diff
@@ -0,0 +1,65 @@
...
+RPM_SPECFILE=rpmbuild/SPECS/fido-device-onboard-rs-$(COMMIT).spec
+RPM_TARBALL=rpmbuild/SOURCES/fido-device-onboard-rs-$(COMMIT).tar.gz
+VENDOR_TARBALL=rpmbuild/SOURCES/fido-device-onboard-rs-$(COMMIT)-vendor.tar.gz
+
+$(RPM_SPECFILE):
+	mkdir -p $(CURDIR)/rpmbuild/SPECS
+	(echo "%global gitversion $(COMMIT)"; git show HEAD:fido-device-onboard.spec) > $(RPM_SPECFILE)
+
+$(RPM_TARBALL):
+	mkdir -p $(CURDIR)/rpmbuild/SOURCES
```
nullr0ute, line 47: Should we add the cargo vendor and bits to make the tar file here too?

runcom: cargo vendor it's right below this

runcom: reason for the split is that fedora doesn't need vendor so when we support fedora in full, we add a condition on fedora itself

fido-device-onboard.spec
nullr0ute: It doesn't read like that here (or in the downloaded spec), and it looks like it has a space. copy from my local spec is ````%{dracutlibdir}/modules.d/52fdo dracut/52fdo/````:

dracut

fido-device-onboard.spec
```diff
@@ -0,0 +1,127 @@
...
+%endif
+BuildRequires: systemd
+
+%description
+%{summary}.
+
+%prep
+%autosetup -n %{name}-rs-%{commit} -p1
+%if 0%{?rhel} && !0%{?eln}
+%cargo_prep -V 1
```
nullr0ute, line 38: Is there a reason we need to -V 1 here and not elsewhere or the reverse?

fido-device-onboard.spec
```diff
@@ -0,0 +1,127 @@
...
+install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-rendezvous-server
+install -D -m 0755 -t %{buildroot}%{_bindir}/fdo target/release/fdo-owner-tool
+install -D -m 0644 -t %{buildroot}%{_unitdir} examples/systemd/*
+install -D -m 0644 -t %{buildroot}%{_docdir}/fdo examples/config/*
+# 52fdo
+install -D -m 0755 -t %{buildroot}%{dracutlibdir}/modules.d/52fdo dracut/52fdo/module-setup.sh
+install -D -m 0755 -t %{buildroot}%{dracutlibdir}/modules.d/52fdo dracut/52fdo/manufacturing-client-generator
+install -D -m 0755 -t %{buildroot}%{dracutlibdir}/modules.d/52fdo dracut/52fdo/manufacturing-client-service
+install -D -m 0755 -t %{buildroot}%{dracutlibdir}/modules.d/52fdo dracut/52fdo/manufacturing-client.service
+
```
nullr0ute, line 60: systemd services need the following lines for each sub package that ships services. add a -n XXX like in the %files for each %XXX

%post
%systemd_post blah.service

%preun
%systemd_preun blah.service

%postun
%systemd_postun_with_restart blah.service

nullr0ute:
Example systemd bits with a rust service here:
https://src.fedoraproject.org/rpms/parsec/blob/rawhide/f/parsec.spec

fido-device-onboard.spec
runcom: this is just for RHEL to use vendored deps

fido-device-onboard.spec
```diff
@@ -0,0 +1,127 @@
...
+%files -n fdo-owner-onboarding-server
+%license LICENSE
+%{_libexecdir}/fdo/fdo-owner-onboarding-server
+%{_docdir}/fdo/owner-onboarding-server.yml
+%{_unitdir}/fdo-owner-onboarding-server.service
+
+%package -n fdo-rendezvous-server
+Summary: FDO Rendezvous Server implementation
+%description -n fdo-rendezvous-server
```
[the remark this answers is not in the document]
runcom:
```
If package scriptlets call other systemd tools, for example systemd-tmpfiles, the package SHOULD declare appropriate dependencies. The %{?systemd_requires} macro is a shortcut to require systemd for the %pre, %post, and %postun scriptlets. Note that those dependencies are not required for the %systemd_{post,preun,postun_with_restart,user_post,user_preun} macros listed above.
```
I think we don't need this but just the macro you suggested below

fido-device-onboard.spec
runcom: I don't know where and how you're getting that :joy: the module is called 52fdo and in my testing it's indeed fdo to pull it in in dracut

runcom: aahhhh that's just the `install` syntax - the space is for install TARGET SRC so doesn't have anything to do with the name too

fido-device-onboard.spec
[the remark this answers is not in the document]
nullr0ute: ````The %{?systemd_requires} macro is a shortcut to require systemd for the %pre, %post, and %postun scriptlets.````

The sentence above states it's a shortcut to require systemd for rpm scriptlets, then you say it's not needed in the following sentence below.

````Note that those dependencies are not required for the %systemd_{post,preun,postun_with_restart,user_post,user_preun} macros listed above.````

fido-device-onboard.spec
runcom: done, notice the manufacturing-client/dracut module don't need this as this is just one-shot in the initramfs

nullr0ute (approved):
LGTM. We can always tweak it as necessary later

fido-device-onboard.spec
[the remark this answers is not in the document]
runcom: > then you say it's not needed in the following sentence below.

those are fedora guidelines not me :smile: and given the above we don't need it as we don't require systemd in pre/post/postun

PR #74, opened 2021-10-18
fix: converge on naming for server(s)

runcom:
Align on naming for server(s) - will need to update the spec and something else later after this lands


puiterwijk:
Note that after that, you'll have to also update the test suite with the new names. After changing the names, do remove your local `target/` directory, so the old files don't keep confusing you with "why does the test suite on my system find the old names and the CI doesn't?"

runcom:
@puiterwijk I believe I've done what you've asked - as for changing the Rust definition from `\*Service\*` to `\*Server\*` I'd rather follow up

PR #76, opened 2021-10-18
Add libfdo-data

puiterwijk:
libfdo-data is a C wrapper around fdo-data-formats, allowing code in
other languages to parse Ownership Vouchers, and possibly other data
formats later on.


puiterwijk:
I explicitly decided to also commit the header file, even though that is auto-generated, so we can ensure that any changes to that are backwards-compatible (given that this is explicitly to be used by other people, `libfdo-data` will be considered stable as of this PR, so please review it as such).

puiterwijk:
> > I think they primarily need the device GUID. I added the protocol version because that was easier during development smile
> 
> Maybe manufacturer/serial number or similar info if it's there, might be useful for customer to identify the device. We can always add that later too.

Yeah, that's a good idea as well, will add that in an upcoming PR.

puiterwijk:
Added a method to get that in #77

PR #94, opened 2021-11-02
fix: add versions of internal dependencies

puiterwijk:
The versions need to be specified for every crate in order to be
publishable to crates.io, until [1] gets implemented.

[1]: https://github.com/rust-lang/rfcs/blob/master/text/2906-cargo-workspace-deduplicate.md


puiterwijk:
Sigh, https://github.com/Keats/tera/issues/689 is hitting us now.
I'm going to see whether he will pull v1.14: if he does, we don't have to do anything.

runcom (approved):

PR #96, opened 2021-11-02
chore: ignore test vouchers with gitleaks

puiterwijk:
This makes it so our ownership vouchers aren't alerted for by gitleaks
when we add the private keys.


nullr0ute (approved):
LGTM

Issue #98, opened 2021-11-03
types: Nonce (/GUID) can be deserialized from an invalid length bstr


puiterwijk:
The Nonce (and thus GUID) type assumes that the data always holds the invariant that it is exactly 16 bytes long.
This is checked by creation, but not by the Deserialize function.
So it is possible for a client to send an invalid length nonce or guid, and cause a `panic!` to our code.

PR #100, opened 2021-11-04
fix: ensure Nonce and Guid are bstr .size 16

puiterwijk:
We make sure that nonce and guid values we receive from "externally"
(i.e. other parties to the protocol, ownership vouchers, etc), are of
the correct length (16 bytes long).
If we receive a value that is not exactly 16 bytes, we will fail to
deserialize.

Fixes: #98

### DIFF
--- a/data-formats/src/types.rs
+++ b/data-formats/src/types.rs
@@ -1,4 +1,9 @@
-use std::{convert::TryInto, ops::Deref, str::FromStr, string::ToString};
+use std::{
+    convert::{TryFrom, TryInto},
+    ops::Deref,
+    str::FromStr,
+    string::ToString,
+};
 
 use aws_nitro_enclaves_cose::crypto::{SigningPrivateKey, SigningPublicKey};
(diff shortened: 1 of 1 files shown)


data-formats/src/types.rs
```diff
@@ -1,4 +1,9 @@
-use std::{convert::TryInto, ops::Deref, str::FromStr, string::ToString};
+use std::{
+    convert::{TryFrom, TryInto},
```
runcom, line 2: silly comment, since rust 1.56 we don't need these anymore but we're leaving them for anyone building with <1.56 right? :thinking:

runcom (approved):

data-formats/src/types.rs
puiterwijk: Yeah, I want to make sure things work with <1.56, since that's not yet in RHEL/Fedora.

runcom: actually it's in fedora but yeah I see

PR #101, opened 2021-11-04
fix: adding golang to devcontainer

jnogol:
Adding Golang to the dev container so Go wrapper tests can work.


puiterwijk (approved):
Thanks!

PR #104, opened 2021-11-05
feat: mfg-client check whether device has active credentials

### LINKED ISSUE #102: manufacturing-client: check whether the device has active FDO creds already
**Goal:** 
The mfg client checks whether the system has credentials in the TPM, if it does, it returns 0

**Acceptance Criteria:**
Check whether the device has active FDO credentials already

jnogol:
Closes #102 

- mfg-client check whether device has active credentials
- Move device credential locations module to util crate


integration-tests/tests/di_diun.rs
```diff
@@ -4,10 +4,12 @@ use std::time::Duration;
 use common::{Binary, LogSide, TestContext};
 
 use anyhow::{bail, Context, Result};
+use serial_test::serial;
```
jnogol, line 7: This (or any other way to execute these tests sequentially) is needed because otherwise the `test_diun` test finds two vouchers instead of one and it fails.

runcom:
This looks good, although we need to open another issue as this case is probably never gonna be it (filesystem in the initramfs) - the other issue is about TPM

jnogol:
I see. I'll need to check with @puiterwijk on how to deal with the TPM.

libfdo-data/Cargo.toml
```diff
@@ -10,7 +10,7 @@ crate-type = ["cdylib"]
 
 [dependencies]
 fdo-data-formats = { path = "../data-formats", version = "0.1.0" }
-
+fdo-util = { path = "../util", version = "0.1.0" }
```
puiterwijk, line 13: Why are you adding this dependency to `libfdo-data`?

integration-tests/tests/di_diun.rs
puiterwijk: Ah, this would be because both tests execute a manufacturing server, and I guess the one from `test_device_credentials_already_active` starts a bit later, so the other one takes both requests.
We should instead make sure that `TestContext` syncs between multiple instances of itself, so that new test servers from other `TestContext`'s get different ports.
I'll try to send a PR for this in a few minutes, because serializing tests is not something I'd like to get into: integration tests can get quite long at some point.

puiterwijk: #105 should fix that, and mean we no longer have to run them serially.

puiterwijk:
@runcom @jnogol this should actually Just Work(R) with TPM based credentials: those would get added to the same `device_credential_location::find()` method, and would be found if they exist 😄

runcom:
Right, do we have a card to track that TPM integration tho? Cause the current issue talks about tpm and it’s not precise at this point

puiterwijk:
Created #106. But basically, because we use the `device_credential_location::find()` method, this will support and device credentials the client-linuxapp supports, automatically.

runcom:
Yup! Lgtm indeed, Patrick feel free to approve and get this

libfdo-data/Cargo.toml
jnogol: That's a very good question I don't have an answer to :laughing:. Blue screen of death moment in my mind, I guess. Removed.

integration-tests/tests/di_diun.rs
jnogol: Cool, removed.

puiterwijk (approved):
LGTM, thanks!

PR #105, opened 2021-11-06
fix: ensure all tests use new server ports

puiterwijk:
Multiple tests in the same test suite should not collide with eachother,
since that would force executing them serially.
We already made sure that all tests run in their own temporary
directory, but we weren't making sure servers don't overlap in bound
ports.
Make sure that within one test suite (i.e. integration-test binary),
ports are chosen sequentially.
This means we can still overlap between different test suites (i.e.
di_diun vs to), but those aren't run in parallel by cargo test by
default anyway.


jnogol (approved):
Tried in my local with #104 and works perfectly, thanks!

PR #113, opened 2021-11-09
fix: use correct lengths for random sizes

puiterwijk:
The random size from the spec is in bits, and we specify it in bytes.
Make sure that we generate and expect the correct number of random
bytes.


data-formats/src/types.rs
```diff
@@ -1190,8 +1190,8 @@ impl<'de> Deserialize<'de> for KexSuite {
 impl KexSuite {
     fn get_ecdh_random_size(&self) -> usize {
         match self {
-            KexSuite::Ecdh256 => 128,
-            KexSuite::Ecdh384 => 384,
+            KexSuite::Ecdh256 => 16,
+            KexSuite::Ecdh384 => 48,
```
runcom, line 1194: damn! good old error :smile:

runcom (approved):

PR #119, opened 2021-11-10
fix: actually enforce tag on COSESign deserialize

puiterwijk:
We were returning an error if the COSESign tag was invalid, but if it
was wrong we were logging an error.
We make sure here that we actually enforce the tags, but then also not
fully crash when we somehow get into the situation where it didn't hold
up.


puiterwijk:
The fun thing is that technically this is a correct failure: those vouchers were invalid.
However, we do want to be able to parse these, to ensure we always keep parsing previously generated ownership vouchers.
In order to accomplish that, I'm working on making vouchers bypass this check.

runcom:
do  you need to update tests here?

puiterwijk:
@runcom yeah, I updated the vouchers used for testing to vouchers that include the tag.

runcom (approved):

Issue #137, opened 2021-11-10
package libfdo_data


runcom:
As a follow up of https://github.com/fedora-iot/fido-device-onboard-rs/pull/70 we need to also ship some sort of `fido-device-onboard-devel` subpkg that contains `libfdo_data.so` (with both %version and %soversion I guess) plus a pkgconfig config (I think). Leving this out of #70 for now.

runcom:
needs to also have support to build the so file with soname https://github.com/rust-lang/cargo/issues/5045 - not sure how to proceed even if we're pretty free to start a downstream soname versioning (see https://docs.fedoraproject.org/en-US/packaging-guidelines/#_shared_libraries)

puiterwijk:
As the last comment says, we could look at using https://crates.io/crates/cargo-c

PR #140, opened 2021-11-12
fix(libfdo-data): build with cargo-c

runcom:
Part of #137 

This just sets the correct SONAME version to the shared library - it wasn't setting anything before this, you can check it out with `scanelf --soname target/debug/libfdo_data.so` 

After this is in, we can think about solving #137 by just installing `libfdo_data.so.0.1.0` and symlinking it to `libfdo_data.so` and `libfdo_data.so.0` (along with packaging the C header file)


puiterwijk (approved):
Awesome, thanks!

puiterwijk:
On the error in `build_and_test`: you should probably make the build script create a symlink?

PR #141, opened 2021-11-12
[data-formats/ownershipvoucher] - Sanitizing manufacturing info

### LINKED ISSUE #54: Manufacturing server and client: check and sanitize Manufacturing Info
Right now, the Manufacturing Info string may contain newlines (for example with the Serial Number one).
This should be sanitized by the client (basically `str.trim()`).
The server should check that this string doesn't include whitespace or other "weird" characters.

jnogol:
Closes #54


data-formats/src/ownershipvoucher.rs
```diff
@@ -441,6 +441,29 @@ impl OwnershipVoucherHeader {
     }
 }
 
+fn check_device_info(device_info: &str) -> Result<bool> {
+    let mut chars = device_info.chars();
+    let result = chars.all(|f| {
+        let acceptable_ascii_range = 32..127;
```
puiterwijk, line 447: Let's make this a global const.

data-formats/src/ownershipvoucher.rs
```diff
@@ -450,7 +473,9 @@ impl Serializable for OwnershipVoucherHeader {
         let cached_guid = contents.get(OwnershipVoucherHeaderIndex::Guid as usize)?;
         let cached_rendezvous_info =
             contents.get(OwnershipVoucherHeaderIndex::RendezvousInfo as usize)?;
-        let cached_device_info = contents.get(OwnershipVoucherHeaderIndex::DeviceInfo as usize)?;
+        let cached_device_info: String =
+            contents.get(OwnershipVoucherHeaderIndex::DeviceInfo as usize)?;
+        if !check_device_info(&cached_device_info).unwrap() { /* return an error? */ };
```
puiterwijk, line 478: You could use `check_device_info(&cached_device_info)?;`, and then the error that that returns will just be propagated up.

data-formats/src/ownershipvoucher.rs
```diff
@@ -441,6 +441,29 @@ impl OwnershipVoucherHeader {
     }
 }
 
+fn check_device_info(device_info: &str) -> Result<bool> {
+    let mut chars = device_info.chars();
+    let result = chars.all(|f| {
+        let acceptable_ascii_range = 32..127;
+        acceptable_ascii_range.contains(&(f as u8))
+    });
+    Ok(result)
```
puiterwijk, line 450: Let's change this function to `-> Result<()>`, and just have something like:
```
if result {
    Err(Error::InconsistentValue("Invalid values in Device Info"))
} else {
    Ok(())
}
```

data-formats/src/ownershipvoucher.rs
jnogol: Fixed.

I wanted to use that but the compiler was complaining, that's why I changed it to that `if` statement. Now I put it back, it's compiling so probably it was an error on my early implementation.

data-formats/src/ownershipvoucher.rs
jnogol: Done.

Question: do you want me to move that global const and [the other one](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/data-formats/src/ownershipvoucher.rs#L58) to the top (after the imports) or do you prefer it to be close to where it's used?

data-formats/src/ownershipvoucher.rs
jnogol: Done

data-formats/src/ownershipvoucher.rs
puiterwijk: Let's move them to the top of the file, just after the imports.

puiterwijk:
Also, perhaps it's an idea to add a `.trim()` call in `OwnershipVoucherHeader::new()`? That way, it is not a problem later on if a newline was added.

jnogol:
If we do that, shall we extend the `ACCEPTABLE_ASCII_RANGE`?

data-formats/src/ownershipvoucher.rs
jnogol: Done.

puiterwijk:
No, I think the current acceptable range is correct. What I meant is that when we create a new OwnershipVoucherHeader, we do a `.trim()`, to remove any newlines that were passed in during creating.
That means we remove newlines at a consistent point, and the actual voucher ends up without the newline in it, and thus passes the current character range.

jnogol:
Got it. Done @puiterwijk

puiterwijk (approved):
LGTM, thanks!

PR #143, opened 2021-11-12
fix(libfdo-data): set soname to the shared library

runcom:
revert cargo-c - add version to the header file


libfdo-data/src/test_common.rs
```diff
@@ -34,6 +34,27 @@ pub fn run_external(script: &str, args: &[&str]) -> Output {
...
+        Err(e) => panic!("Failed to remove libfdo_data.so symlink {:?}", e)
+
+    }
+    std::os::unix::fs::symlink(
+        format!(
+            "{}/target/debug/libfdo_data.so",
+            std::env::var("PWD").unwrap()
+        ),
+        format!(
+            "{}/target/debug/libfdo_data.so.0",
```
puiterwijk, line 52: Maybe make sure that this `.0` is automatically provided by soname version?

runcom: do we want to shell out to grab that? perhaps cargo has the env in tests?

libfdo-data/src/test_common.rs
```diff
@@ -34,6 +34,27 @@ pub fn run_external(script: &str, args: &[&str]) -> Output {
...
+    )) {
+        Ok(_) => (),
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => (),
+        Err(e) => panic!("Failed to remove libfdo_data.so symlink {:?}", e)
+
+    }
+    std::os::unix::fs::symlink(
+        format!(
+            "{}/target/debug/libfdo_data.so",
+            std::env::var("PWD").unwrap()
```
puiterwijk, line 49: So, I know how ugly it can be to find things, but maybe let's use `env!("CARGO_MANIFEST_DIR")`, since `PWD` feels very fragile?

libfdo-data/src/test_common.rs
puiterwijk: Cargo should be setting `CARGO_VERSION_MAJOR` as part of the build, even for tests I believe?

runcom: oh yeah it works

libfdo-data/src/test_common.rs
runcom: done

puiterwijk (approved):

Issue #81, opened 2021-10-22
Client: ensure that we stop TO2 if to1d was invalid


puiterwijk:
If the to1d was not signed by the valid owner of the OV, we need to stop the full TO2 protocol.

PR #145, opened 2021-11-15
feat: cancel to2 if to1d was invalid

puiterwijk:
If the to1d we received as part of the TO1 protocol is invalid, we need
to stop the TO2 that we were performing with that same to1d document.

Fixes: #81


jnogol (approved):
LGTM

PR #146, opened 2021-11-15
feat: perform report to rendezvous periodically

### LINKED ISSUE #92: Owner onboarding service: take care of reporting to rendezvous
The Owner Onboarding Service should perform TO0 regularly for the ownership vouchers under its control, to make sure that the owner doesn't have to do so manually.

runcom:
Fix #92

### DIFF
--- a/integration-tests/tests/common/mod.rs
+++ b/integration-tests/tests/common/mod.rs
@@ -739,6 +739,11 @@ impl<'a> TestServerConfigurator<'a> {
                     "bind",
                     &format!("127.0.0.1:{}", self.server_number.server_port().unwrap()),
                 );
+                cfg.insert("owner_port", &self.server_number.server_port().unwrap());
+                cfg.insert(
+                    "config_dir",
+                    &self.test_context.runner_path(&self.server_number),
+                );
                 // TODO: Insert more defaults
 
                 context_configurator(cfg)
--- a/integration-tests/tests/to.rs
+++ b/integration-tests/tests/to.rs
@@ -22,10 +22,13 @@ async fn test_to() -> Result<()> {
     let owner_onboarding_server = ctx
         .start_test_server(
             Binary::OwnerOnboardingServer,
-            |cfg| Ok(cfg.prepare_config_file(None, |_| Ok(()))?),
+            |cfg| {
+                cfg.prepare_config_file(Some("owner-addresses.yml"), |_| Ok(()))?;
+                Ok(cfg.prepare_config_file(None, |_| Ok(()))?)
+            },
             |_| Ok(()),
         )
-        .context("Error creating rendezvous server")?;
+        .context("Error creating owner server")?;
     ctx.wait_until_servers_ready()
         .await
         .context("Error waiting for servers to start")?;
@@ -33,7 +36,6 @@ async fn test_to() -> Result<()> {
     let dc_path = ctx.testpath().join("testdevice.dc");
     let ov_path = ctx.testpath().join("testdevice.ov");
     let rendezvous_info_path = ctx.testpath().join("rendezvous-info.yml");
-    let owner_address_path = ctx.testpath().join("owner-addresses.yml");
     let key_path = ctx.keys_path();
 
     L.l("Generating configuration files");
@@ -42,14 +44,6 @@ async fn test_to() -> Result<()> {
         Ok(())
     })
     .context("Error generating rendezvous-info.yml")?;
-    ctx.generate_config_file(&owner_address_path, "owner-addresses.yml", |cfg| {
-        cfg.insert(
-            "owner_port",
-            &owner_onboarding_server.server_port().unwrap(),
-        );
-        Ok(())
-    })
-    .context("Error generating owner-addresses.yml")?;
 
     let owner_output = ctx
         .run_owner_tool(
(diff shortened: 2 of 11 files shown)


integration-tests/tests/common/mod.rs
```diff
@@ -723,6 +723,11 @@ impl<'a> TestServerConfigurator<'a> {
                     "bind",
                     &format!("127.0.0.1:{}", self.server_number.server_port().unwrap()),
                 );
+                cfg.insert("owner_port", &self.server_number.server_port().unwrap());
```
puiterwijk, line 726: Maybe call this server_port to be consistent?

integration-tests/tests/to.rs
```diff
@@ -22,18 +22,20 @@ async fn test_to() -> Result<()> {
...
             Binary::OwnerOnboardingServer,
-            |cfg| Ok(cfg.prepare_config_file(None, |_| Ok(()))?),
+            |cfg| {
+                cfg.prepare_config_file(Some("owner-addresses.yml"), |_| Ok(()))?;
+                Ok(cfg.prepare_config_file(None, |_| Ok(()))?)
+            },
             |_| Ok(()),
         )
-        .context("Error creating rendezvous server")?;
+        .context("Error creating owner server")?;
```
puiterwijk, line 31: Nice find, that was a copy-paste error 😄

owner-onboarding-server/src/main.rs
```diff
@@ -19,15 +24,19 @@ use tokio::signal::unix::{signal, SignalKind};
...
+    types::{Guid, TO2AddressEntry},
 };
 use fdo_store::{Store, StoreDriver};
 use fdo_util::servers::{settings_for, OwnershipVoucherStoreMetadataKey};
 
 mod handlers;
 mod serviceinfo;
 
-struct OwnerServiceUD {
+pub struct OwnerServiceUD {
```
puiterwijk, line 39: Let's make this a `pub(crate)`?

runcom: this is wrong, I'll drop it - and the handler for report to rendezvous is still work in progress (pending some more learning around warp filters :))

owner-onboarding-server/src/main.rs
```diff
@@ -79,13 +90,129 @@ struct Settings {
...
+    ft.lt(
+        &fdo_store::MetadataKey::Local(OwnershipVoucherStoreMetadataKey::To0AcceptOwnerWaitSeconds),
+        chrono::Local::now().timestamp(),
+    );
+
+    let ov_iter = ft.query().await?;
+    if let Some(ovs) = ov_iter {
+        for ov in ovs {
+            let wait_seconds =
+                report_ov_to_rendezvous(&ov, &udt.owner_addresses, &udt.owner_key).await?;
```
puiterwijk, line 118: Perhaps make this a separate function, and log but not fail on error?
Otherwise, if one of the OV's fails, we also stop reporting the others.

integration-tests/tests/common/mod.rs
runcom: uhm, not sure I understand :thinking:

owner-onboarding-server/src/main.rs
runcom: oh right, I'll just match here and log then

owner-onboarding-server/src/main.rs
```diff
@@ -266,10 +423,18 @@ async fn main() -> Result<()> {
         handlers::done,
     );
 
+    // TODO(runcom): enable this conditionally from configuration
```
puiterwijk, line 426: It's significantly easier to instead always add the handler, but make the route itself just return an error if it is disabled (and make it disabled by default).
It's... tricky to create a `let routes =` where sometimes a route is there and sometimes it isn't.

owner-onboarding-server/src/main.rs
```diff
@@ -304,3 +469,111 @@ async fn main() -> Result<()> {
 
     Ok(())
 }
+
+// MOVE TO TYPES?
```
puiterwijk, line 473: Yes 😃

integration-tests/tests/common/mod.rs
puiterwijk: Basically, this is inserted always, also with rendezvous server.
Maybe just make this call `cfg.insert("server_port", &self.server_number.server_port().unwrap());`?

owner-onboarding-server/src/main.rs
puiterwijk: Ah, okay.

owner-onboarding-server/src/main.rs
puiterwijk: For an example here, see the DIUN implementation.

integration-tests/tests/common/mod.rs
runcom: > Basically, this is inserted always, also with rendezvous server.

is it? I see that only in rendezvous-info.yml in the tests :thinking:

owner-onboarding-server/src/main.rs
runcom: I’ve pushed again by just passing a Boolean that I’ll later set - I’ll revisit this PR again tomorrow morning

runcom: Right looking at the DIUN example I’m doing mostly the same except I’m returning 404 directly - no need to expose this at all even if it’s disabled

owner-onboarding-server/src/main.rs
runcom: done!

runcom:
now ready to review

integration-tests/tests/common/mod.rs
runcom: I think it's inserted with specific ports when calling generate_config_file so not within a server creation where it can inject its own port

puiterwijk (approved):
Thank you very much for this work!

Issue #130, opened 2021-11-10
Change default loglevel to `info`


puiterwijk:
Right now, the default log level is unset.
We should make sure that this by default is set to `info`, so that informative things are logged.

PR #147, opened 2021-11-15
feat: default log level to info

puiterwijk:
Make sure that by default we log at least at the INFO level, since that
is useful pretty important information.

Fixes: #130


runcom (approved):

puiterwijk:
@Mergifyio refresh

PR #157, opened 2021-11-18
fix(spec): use forgesource

runcom:

### DIFF
--- a/fido-device-onboard.spec
+++ b/fido-device-onboard.spec
@@ -13,7 +13,7 @@ Summary:        An implementation of the FIDO Device Onboard Specification writt
 
 License:        BSD 3
 URL:            %{forgeurl}
-Source:         %{url}/archive/%{commit}/%{name}-rs-%{commit}.tar.gz
+Source:         %{forgesource}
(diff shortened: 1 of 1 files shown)


fido-device-onboard.spec
```diff
@@ -13,7 +13,7 @@ Summary:        An implementation of the FIDO Device Onboard Specification writt
 
 License:        BSD 3
 URL:            %{forgeurl}
-Source:         %{url}/archive/%{commit}/%{name}-rs-%{commit}.tar.gz
+Source:         %{forgesource}
```
nullr0ute, line 16: You still need a tar file in that URL, I don't see a forgesource defined anywhere in the source tree using grep.

runcom: https://fedoraproject.org/wiki/Forge-hosted_projects_packaging_automation

nullr0ute: For reference the docs are at: https://docs.fedoraproject.org/en-US/packaging-guidelines/SourceURL/#_using_forges_hosted_revision_control

nullr0ute (approved):
LGTM

PR #158, opened 2021-11-18
fix(spec): use forgesetup

runcom:
Hopefully the last PR around the spec file @nullr0ute :angel: 

This allows us to keep building with make rpm just fine and just change
the vendor tar gz downstream variable to version instead of commit.


nullr0ute (approved):
Good to see we tested #157 end to end then :1st_place_medal: 
LGTM

Issue #155, opened 2021-11-17
ssh key module: create .ssh and authorized_keys with the ssh_user property


runcom:
In `.ssh/authorized_keys` we limit ourselves to just append to it leaving the current uid:gid as owner - however, on a system where there's no `.ssh/authorized_keys` the onboarding process creates `.ssh` and `authorized_keys` as `root`, ignoring the `user` from the ssh_user module - that results in a system where I cannot login as the auth.._key file is not readable by anybody but root. We should just set file/dir ownership on creation to whatever specified in the ssh module

PR #159, opened 2021-11-18
fix(serviceinfo): install ssh key with correct ownership

runcom:
Fix https://github.com/fedora-iot/fido-device-onboard-rs/issues/155


runcom:
> It mostly looks fine, does the new nix crate allow us to replace the custom set_perm_mode function that sets the ssh dir/key dir permissions?

I'll check that out - currently fixing tests too

puiterwijk (approved):

PR #162, opened 2021-11-19
fix(spec): build with version if commit undefined

runcom:
This allows us to have the same upstream and downstream spec file.
commit is needed for mainly dev and snapshots, version is what we really
use once we build downstream in Fedora/RHEL. `commit` is either defined in dev using forgemeta and Makefile or not set downstream (cause there we usually ship tagged versions)


nullr0ute (approved):
LGTM

Issue #84, opened 2021-10-27
paths in config should be absolute (and checked)


runcom:
Running the servers with systemd and having relative paths in configs (for things like the OV folder or sessions or keys) doesn't error out, we need to check that and mandate absolute paths

PR #165, opened 2021-12-02
fix: check paths in configs are absolute

runcom:
Fix https://github.com/fedora-iot/fido-device-onboard-rs/issues/84

### DIFF
--- a/integration-tests/tests/di_diun.rs
+++ b/integration-tests/tests/di_diun.rs
@@ -14,7 +14,12 @@ async fn test_diun() -> Result<()> {
     let mfg_server = ctx
         .start_test_server(
             Binary::ManufacturingServer,
-            |cfg| Ok(cfg.prepare_config_file(None, |_| Ok(()))?),
+            |cfg| {
+                cfg.prepare_config_file(Some("rendezvous-info.yml"), |ctx| {
+                    Ok(ctx.insert("rendezvous_port", &1337))
+                })?;
+                Ok(cfg.prepare_config_file(None, |_| Ok(()))?)
+            },
             |_| Ok(()),
         )
         .context("Error creating manufacturing server")?;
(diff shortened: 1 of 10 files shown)


integration-tests/tests/di_diun.rs
```diff
@@ -14,7 +14,12 @@ async fn test_diun() -> Result<()> {
     let mfg_server = ctx
         .start_test_server(
             Binary::ManufacturingServer,
-            |cfg| Ok(cfg.prepare_config_file(None, |_| Ok(()))?),
+            |cfg| {
+                cfg.prepare_config_file(Some("rendezvous-info.yml"), |ctx| {
+                    Ok(ctx.insert("rendezvous_port", &1337))
```
runcom, line 19: this is just a dummy port - it's not needed for the test anyway and no reason to bring up the rendezvous either - let me know if anybody feels otherwise

manufacturing-server/src/main.rs
```diff
@@ -165,21 +167,15 @@ impl TryFrom<DiunSettings> for DiunConfiguration {
...
-        Ok(c) => c,
-        // backwards compatibility for tests...
-        Err(_) => return Ok(RendezvousInfo::new(Vec::new())),
-    };
-    let mut info: Vec<RendezvousDirective> = Vec::new();
-
+fn load_rendezvous_info(path: &AbsolutePathBuf) -> Result<RendezvousInfo> {
     let value: Value =
-        serde_yaml::from_slice(&contents).context("Error parsing rendezvous info")?;
+        serde_yaml::from_slice(&fs::read(path)?).context("Error parsing rendezvous info")?;
```
puiterwijk, line 172: Let's do an `let contents = fs::read(path).context("Failed to read rendezvous info");`, and then pass `&contents` in here: basically make sure that the file read returns a different error string than failing to parse.

puiterwijk (commented):
Overall, quite nice. Just one small request for fix :).

manufacturing-server/src/main.rs
runcom: fixed

puiterwijk (approved):
LGTM :shipit: thanks!

PR #166, opened 2021-12-06
Implement parsing multiple Ownership Vouchers, and add libfdo-data-go

### LINKED ISSUE #148: libfdo_data to support multiple OVs parsing
For now, the parsing support only 1 OV at a time, when trying to parse a batch of OVs this won't work.

puiterwijk:
This patchset fixes #148, by supporting parsing multiple Ownership Voucher parsing, and adding support for this to libfdo-data.
It also adds a libfdo-data-go Go wrapper, which users can then use if they use libfdo-data from Go.


puiterwijk:
The stability impact is caused by adding new functions to libfdo-data.

libfdo-data-go/ownershipvoucher.go
```diff
@@ -0,0 +1,86 @@
...
+	if ov.should_free {
+		C.fdo_ownershipvoucher_free(ov.inner)
+	}
+}
+
+func (ov *OwnershipVoucher) GetProtocolVersion() uint32 {
+	return uint32(C.fdo_ownershipvoucher_header_get_protocol_version(ov.inner))
+}
+
+func (ov *OwnershipVoucher) GetGUID() *FDOString {
```
puiterwijk, line 80: I'm going to improve the String handling here

libfdo-data-go/ownershipvoucher.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (ovl *OwnershipVoucherList) Free() {
+	C.fdo_ownershipvoucher_list_free(ovl.inner)
+}
+
+func (ovl *OwnershipVoucherList) Len() int {
+	return int(C.fdo_ownershipvoucher_list_len(ovl.inner))
+}
+
+func (ovl *OwnershipVoucherList) GetVoucher(index int) *OwnershipVoucher {
```
runcom, line 36: I would do this as:

```go
func (ovl *OwnershipVoucherList) GetVoucher(index int) (*OwnershipVoucher, error) {
```

err in golang is always checked anyway and you can return a Not Found in the `== nil` case

runcom (approved):
apart from that interface change in `GetVoucher` - lgtm

puiterwijk:
@runcom I changed the API to your suggestion (thanks!), and added a `go doc` output for CI labeling, please take a look.

PR #169, opened 2021-12-09
feat: add non-interoperable KDF

puiterwijk:
This adds the ability to add a non-interoperable KDF implementation.
This should never be used in production environments, as it will be not
interoperable with other implementations, or this implementation with
the interoperable KDF.

When the non-interoperable KDF is enabled, the client and server will
send a header to indicate this, so it becomes possible for the other
side to dynamically fall back to the non-interoperable KDF when needed.
This is work for a future patch, but this ensures that such a patch will
have the required information.

### DIFF
--- a/client-linuxapp/src/main.rs
+++ b/client-linuxapp/src/main.rs
@@ -460,6 +460,10 @@ fn sleep_between_retries(rv_entry_delay: u32) {
 async fn main() -> Result<()> {
     fdo_http_wrapper::init_logging();
 
+    if !fdo_data_formats::INTEROPERABLE_KDF && std::env::var("ALLOW_NONINTEROPABLE_KDF").is_err() {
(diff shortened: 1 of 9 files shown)


puiterwijk:
In order to get a build that's FIPS-compliant (but non-interoperable!) with this patch, build like this: `cargo build --features "openssl-kdf/deny_custom,fdo-data-formats/use_noninteroperable_kdf"`

runcom:
> In order to get a build that's FIPS-compliant (but non-interoperable!) with this patch, build like this: `cargo build --features "openssl-kdf/deny_custom,fdo-data-formats/use_noninteroperable_kdf"`

I'll test a build with this - otherwise it looks good to me :+1:

runcom (approved):

runcom (approved):

PR #174, opened 2021-12-10
Remove todos, limit panics, and check for them

puiterwijk:


http-wrapper/src/server.rs
```diff
@@ -20,12 +20,13 @@ pub struct SessionWithStore {
...
 type SessionStoreT = Arc<SessionStore>;
 
+#[derive(Debug, Clone, Copy)]
 #[non_exhaustive]
 pub enum SessionStoreMetadataKey {}
 
 impl MetadataLocalKey for SessionStoreMetadataKey {
     fn to_key(&self) -> &'static str {
-        todo!()
+        match *self {}
```
runcom, line 29: oh neat what is this?

runcom: does just nothing and returns ""?

runcom (approved):
just a curiosity comment, lgtm otherwise

http-wrapper/src/server.rs
puiterwijk: No, the `*self` makes it go from a `&SessionStoreMetadataKey` to a `SessionStoreMetadataKey`, which we can do without performance impact because it derives `Copy`.
This dereferencing is done because a pointer to an enum can be something (empty enums are often used in FFI to represent an abstract C type).

After that, we match on `SessionStoreMetadataKey`, and the compiler is smart enough to realise that there's no values in that enum, so it will never match.
So then the compiler realises there's no chance at all that this function ever gets called.
So no, it doesn't return an empty string (rust would require you to be explicit for that sort of stuff and doesn't implicitly return things), the compiler just realises that this never matches, so that the function can never be called.

runcom: uhm I see :thinking:

Issue #129, opened 2021-11-10
CI: Check if we can use the standard images for CI and devcontainer


puiterwijk:
We should check, now that we do our own KDF and no longer depend on Fedora/RHEL patches in OpenSSL, whether we can use the standard images for GitHub actions and devcontainer.
This would probably speed up CI and devcontainer building significantly.

Fedora/CentOS tests should arrive as part of Packit.

puiterwijk:
This has been done a while ago.

puiterwijk:
Actually, no. We moved CI, devcontainer is not moved yet.

PR #177, opened 2021-12-10
feat: move to standard devcontainers image

puiterwijk:
Fixes: #129


puiterwijk:
This should make the startup time for the devcontainer significantly faster.

puiterwijk:
And we can't do this for now anymore because of the use of TPM.

PR #182, opened 2021-12-11
fix(spec): Version has to come before forgemeta

runcom:


nullr0ute (commented):
Rather than just moving Version up can we move %forgemeta down to below Source?

runcom:
> Rather than just moving Version up can we move %forgemeta down to below Source?

unfortunately we can't - see osbuild doing the same https://github.com/osbuild/osbuild/blob/main/osbuild.spec (and many others)

puiterwijk (approved):

PR #186, opened 2022-01-05
Implement file and command serviceinfo modules

puiterwijk:
This implements a base `binaryfile` and `command` serviceinfo module set.
Both of them are currently quite hardwired on the server, but that can be changed later when the service info API gets implemented.

### DIFF
--- a/client-linuxapp/src/serviceinfo.rs
+++ b/client-linuxapp/src/serviceinfo.rs
@@ -1,13 +1,19 @@
-use std::path::Path;
 use std::process::Command;
+use std::{
+    collections::HashSet,
+    fs::{File, Permissions},
+    io::Write,
+    path::Path,
+};
 use std::{env, fs};
 use std::{os::unix::fs::PermissionsExt, path::PathBuf};
 
 use anyhow::{anyhow, bail, Context, Result};
 
 use fdo_data_formats::{
+    constants::HashType,
     messages::to2::{DeviceServiceInfo, OwnerServiceInfo},
-    types::{CborSimpleTypeExt, ServiceInfo},
+    types::{CborSimpleTypeExt, Hash, ServiceInfo},
 };
 use fdo_http_wrapper::client::{RequestResult, ServiceClient};
 
@@ -18,6 +24,7 @@ fn find_available_modules() -> Result<Vec<String>> {
         // These modules are always here
         "devmod".to_string(),
         "sshkey".to_string(),
+        "binaryfile".to_string(),
     ];
 
     // See if we add RHSM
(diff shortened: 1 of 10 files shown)


client-linuxapp/src/serviceinfo.rs
```diff
@@ -159,11 +288,138 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo) -> Result<()> {
...
+                        binary_file_in_progress.mode
+                    );
+                }
+                binary_file_in_progress.mode =
+                    Some(value.as_u32().context("Error parsing binaryfile mode")?);
+            } else if key.starts_with("sha-") {
+                let sha_type = key.split('-').nth(1).unwrap();
+                let sha_value = value
+                    .as_bytes()
+                    .with_context(|| format!("Error parsing binaryfile sha-{} value", sha_type))?;
```
runcom, line 342: `s/binaryfile/binary file/g`?

integration-tests/templates/owner-onboarding-server.yml.j2
```diff
@@ -10,8 +10,29 @@ owner_addresses_path: {{ config_dir }}/owner-addresses.yml
 report_to_rendezvous_endpoint_enabled: true
 bind: {{ bind }}
 service_info:
-  rhsm_organization_id: 6454383
+  rhsm_organization_id: 123456
   rhsm_activation_key: SDOkey
   rhsm_run_insights: true
   sshkey_user: {{ user }}
   sshkey_key: "testkey"
+  files:
```
runcom, line 18: https://github.com/coreos/ignition/pull/1285 ugh

puiterwijk: Hmm, okay. So you want me to rename this?

client-linuxapp/src/serviceinfo.rs
puiterwijk: Sure. Will push the fix after your first review pass finishes.

integration-tests/templates/owner-onboarding-server.yml.j2
runcom: oh no, just wondering when/if we can use something like that to avoid reimplementing file handling

puiterwijk: So, the thought here is that right now we don't have any other useful serviceinfo modules.
This one together with the process one could be used to kickstart that, by for example putting in a file with an ignition or ansible file, and then have the command module call ignition/ansible.
i.e. this is just a basic one in order to allow for those types of use cases.

runcom (approved):

PR #188, opened 2022-01-12
chore(systemd): always run the fdo client

runcom:
Drop the condition on the installer having written credentials under
/boot - have the dracut module write an env file to read that on startup
or default to empty.


examples/systemd/fdo-client-linuxapp.service
```diff
@@ -5,7 +5,7 @@ After=network-online.target
```
puiterwijk, line 5: You did not actually remove the ConditionPathExists

runcom: done

puiterwijk (approved):
LGTM

examples/systemd/fdo-client-linuxapp.service
```diff
@@ -1,11 +1,10 @@
...
-ConditionPathExists=/boot/device-credentials
 After=network-online.target
 
 [Service]
 Type=oneshot
-Environment=DEVICE_CREDENTIAL=/boot/device-credentials
+EnvironmentFile=/boot/fdo-client-env
 Environment=LOG_LEVEL=info
 ExecStart=/usr/libexec/fdo/fdo-client-linuxapp
 ExecStartPost=/usr/bin/mv /boot/device-credentials /etc/device-credentials
```
runcom, line 10: we should drop this one also I think or make it ok to fail

PR #190, opened 2022-01-14
chore(systemd): mv creds may fail

runcom:
mv may fail if creds aren't there - take that into account.


runcom:
@puiterwijk why is clippy complaining here?

puiterwijk:
@runcom ah, that's probably a new lint (which does make sense). I'll send a PR to fix this in a minute.

puiterwijk:
See #191

runcom:
should pass now then :)

puiterwijk (approved):

PR #192, opened 2022-01-17
Move to FDO version 1.1

puiterwijk:
This patch makes it possible to implement different versions of the
spec, by moving messages into a version-specific submodule.


client-linuxapp/src/main.rs
```diff
@@ -498,6 +501,16 @@ async fn main() -> Result<()> {
         log::info!("Device credential deactivated, skipping Device Onboarding");
         return Ok(());
     }
+    if dc.protocol_version() != ProtocolVersion::Version1_1 {
+        log::info!(
```
runcom, line 505: do we need the log here? doesn't bailing suffice as it's going to be logged anyway?

runcom:
lgtm apart from a nit-picking comment :)

runcom (approved):

client-linuxapp/src/main.rs
puiterwijk: Yep.

PR #196, opened 2022-02-04
chore(examples/config): fix manufacturing server config

runcom:
change store driver to Directory, we dropped InMemory
a while ago


puiterwijk (approved):
Thanks!

Issue #203, opened 2022-02-09
fdo-linuxclient-app doesn't look for /etc/device-credentials


mangelajo:
Will only look if `DEVICE_CREDENTIAL=/etc/device-credentials` is set, but considering that the manufacturing client uses /etc/device-credentials as a default, I'd propose looking for that file (with lower preference the DEVICE_CREDENTIALS or qemu injected file).

PR #204, opened 2022-02-09
default DEVICE_CREDENTIAL to /etc/device-credentials

mangelajo:
This will match fdo-linuxclient-app to the fdo-manufacturing-client
behavior.

Fixes-Issue: #203

Signed-off-by: Miguel Angel Ajo <majopela@redhat.com>

### DIFF
--- a/util/src/device_credential_locations.rs
+++ b/util/src/device_credential_locations.rs
@@ -16,6 +16,10 @@ pub fn find() -> Option<Result<Box<dyn UsableDeviceCredentialLocation>>> {
         Box::new(FileSystemPathEnv {
             env_var: "DEVICE_CREDENTIAL".to_string(),
         }),
+        Box::new(FileSystemPath {
+            path: "/etc/device-credentials".to_string(),
+            deactivation_method: DeactivationMethod::Deactivate,
+        }),
     ];
 
     for devcredloc in device_credential_locations {


puiterwijk:
That's not the way we prefer to add additional paths.
This is the `FileSystemPathEnv` location provider, which should only be for reading environment variables.
In order to also look at `/etc/device-credentials`, you'd add a `FileSystemPath` instance at [`pub fn find()`](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/util/src/device_credential_locations.rs#L10)

puiterwijk (changes requested):
The intention is that the `find()` method has a list of places where to search for device credentials, and that each possible place is either "yes, it's there", or "no, not there".
So having the environment variable say "yes, it's there", all the time, is not ideal.

mangelajo:
hey thanks @puiterwijk yes, that's what I understood from the code.

What I found missing then was the ability to define a deactivation method:

https://github.com/fedora-iot/fido-device-onboard-rs/pull/204/files#diff-a6defb265b6577f2e084ab6d98f948240cc7734140f8516ea63b091a3ba9cba0R11

In my first version of the patch I used what you suggested, but I found the ability to specify deactivation lacking.

What's your opinion on this if I switch back to that form?, should we capture the deactivation method from environment in that case? (I understand that the qemu provided is not mutable, but a filesystem one would be...).

I followed this path because it's the last priority element, and did want to provide flexibility on deactivation, but let's decide how to handle that detail and I can update the PR.

puiterwijk:
How about adding this to `fn find()`:
```
        Box::new(FileSystemPath {
            path: "/etc/device-credentials".to_string(),
            deactivation_method: DeactivationMethod::Deactivate,
        }),
```

mangelajo:
Yes, that was my first version, but then there is no control on deactivation. Is it mandatory to always deactivate? what if we want to support ownership transfer later?

mangelajo:
It's fine for me, I can update the patch to that :) only wanted to share my line of thought, and why I ended doing it via environment.

puiterwijk (approved):

puiterwijk:
Thanks for the patch!
Your commit message will need a "type" prefix, possibly `feature: `?

mangelajo:
Sorry for the slow update, I haver rebased on main, and updated the commit message.

runcom:
lgtm

puiterwijk:
@Mergifyio rebase

puiterwijk (dismissed):

7flying:
@mangelajo ci complains about the commit message, can you amend "feature:" to "feat:"?

mangelajo:
> @mangelajo ci complains about the commit message, can you amend "feature:" to "feat:"?

sure thing, 1 sec

mangelajo:
@7flying, the commit should be good now 🤞; thanks for the ping! 👍

7flying (approved):
LGTM
Thanks for your contribution!

PR #207, opened 2022-02-10
fix(owner-onboarding-server): Re-run TO0 before it expires

### LINKED ISSUE #193: Owner Onboarding Server: make sure to re-run TO0 before it is actually expired
We should add a window where "if the registration expires in the next N minutes, we re-execute TO0", so there's no moment the device is not registered at the rendezvous server.

7flying:
Fixes #193 
- We re-run TO0 before it expires based on a window.
- `registration_time` + `re_registration_window` are now configured on the .yml.
  -  The sample `owner-onboarding-server.yml` has been updated with these new parameters.
  -  We have default values for the `registration_time` + `re_registration_window`.

### DIFF
--- a/owner-onboarding-server/src/main.rs
+++ b/owner-onboarding-server/src/main.rs
@@ -416,9 +416,7 @@ async fn main() -> Result<()> {
     };
 
     // Voucher registration times
-    let str_registration_period = settings.registration_period;
-    let str_re_registration_window = settings.re_registration_window;
-    let registration_period = match str_registration_period {
+    let registration_period = match settings.registration_period {
         Some(value) => {
             let tmp = value
                 .parse::<u32>()
@@ -430,20 +428,13 @@ async fn main() -> Result<()> {
         }
         None => DEFAULT_REGISTRATION_PERIOD,
     };
-    let re_registration_window = match str_re_registration_window {
+    let re_registration_window = match settings.re_registration_window {
         Some(value) => {
             let tmp = value
                 .parse::<u32>()
                 .context("must provide a valid re_registration_window")?;
             if tmp == 0 {
                 bail!("re_registration_window cannot be 0");
-            }
-            if tmp >= registration_period {
-                bail!(
-                    "re_registration_window ({}) must be smaller than registration_period ({})",
-                    tmp,
-                    registration_period
-                );
             }
             tmp
         }


owner-onboarding-server/src/main.rs
```diff
@@ -95,6 +100,10 @@ struct Settings {
     owner_addresses_path: AbsolutePathBuf,
 
     report_to_rendezvous_endpoint_enabled: bool,
+
+    // in seconds
+    registration_period: Option<String>,
```
puiterwijk, line 105: Is there a specific reason you don't do `Option<u32>`?

owner-onboarding-server/src/main.rs
```diff
@@ -244,6 +305,7 @@ async fn perform_maintenance(udt: OwnerServiceUDT) -> std::result::Result<(), &'
 
         #[allow(unused_must_use)]
         let (ov_res, ses_res, rtr_res) = tokio::join!(ov_maint, ses_maint, rtr_maint);
+        let win_res = check_registration_window(udt.clone()).await;
```
puiterwijk, line 308: Having this here would mean that this only gets executed after the maintenance or serving job has stopped. This code should probably be executed in the `perform_maintenance` method, and have that take the window into account.

owner-onboarding-server/src/main.rs
```diff
@@ -346,6 +415,48 @@ async fn main() -> Result<()> {
             .context("Error converting owner public key to PK")?
     };
 
+    // Voucher registration times
+    let str_registration_period = settings.registration_period;
+    let str_re_registration_window = settings.re_registration_window;
+    let registration_period = match str_registration_period {
```
puiterwijk, line 421: You could do a `let registration_period = match settings.registration_period {`, which lets you get around the `let str_...`.

owner-onboarding-server/src/main.rs
```diff
@@ -346,6 +415,48 @@ async fn main() -> Result<()> {
...
+        Some(value) => {
+            let tmp = value
+                .parse::<u32>()
```
puiterwijk, line 424: If you make the type `Option<u32>`, you don't have to parse and handle parse errors anymore.

owner-onboarding-server/src/main.rs
```diff
@@ -346,6 +415,48 @@ async fn main() -> Result<()> {
...
+    };
+    let re_registration_window = match str_re_registration_window {
+        Some(value) => {
+            let tmp = value
+                .parse::<u32>()
+                .context("must provide a valid re_registration_window")?;
+            if tmp == 0 {
+                bail!("re_registration_window cannot be 0");
+            }
+            if tmp >= registration_period {
```
puiterwijk, line 441: This check seems to be duplicated, as it's also executed later on.

owner-onboarding-server/src/main.rs
7flying: Yes, to the best of my knowledge serde doesn't have a way to check that u32 must be positive while loading the settings from the file, even if we are requesting a u32, thereby my tests show, that if I put a minus something it will be wrapped to a random positive integer (on target debug and release).

I haven't found a way to catch when serde does this. There are some crates to allow these types of checks but they didn't look stable enough and we have a way to check this without external crates.

owner-onboarding-server/src/main.rs
7flying: Related to the first comment, in this way I check that the number is not negative while I parse it.

owner-onboarding-server/src/main.rs
7flying: Yes, since the `registration_window` and `re_registration_window` are not compulsory fields, a user may choose to config one, both or neither of them. In the case that only one of them is configured the other one will get the default value, thereby we need to check again that  window < registration_period.

owner-onboarding-server/src/main.rs
puiterwijk: Ah, I would've expected signed integers to cause an error. That's unfortunate, but fair enough then.

owner-onboarding-server/src/main.rs
puiterwijk: Yep, fair enough then.

owner-onboarding-server/src/main.rs
puiterwijk: Yep, I was initially going to type it would be best to move this down, and then I saw it was also there. So maybe just only use that one, and drop this one?

owner-onboarding-server/src/main.rs
7flying: I don't know what you mean by 'after the job has stopped'. The `check_registration_window` is called inside the `perform_maintenance` method, as you ask. It simply waits until possible OV registration plus session maintenance has been made.
The window is taken into account inside the `check_registration_window` when we query `fdo_store`.

owner-onboarding-server/src/main.rs
7flying: Will do.

owner-onboarding-server/src/main.rs
7flying: Will do, good catch.

owner-onboarding-server/src/main.rs
puiterwijk: I'm sorry, this was me mis-remembering how many `join!` calls there were in the code: I thought the only one was in the `fn main()`.

owner-onboarding-server/src/main.rs
```diff
@@ -141,10 +157,58 @@ async fn report_to_rendezvous(udt: OwnerServiceUDT) -> Result<()> {
     Ok(())
 }
 
+async fn check_registration_window(udt: OwnerServiceUDT) -> Result<()> {
```
puiterwijk, line 160: We've discussed that this got pulled out separately because the query mechanism doesn't quite do what we'd want.
I'm going to look at improving and/or documenting that, and then get back to this shortly.

Issue #153, opened 2021-11-16
Implement Service Info API client


puiterwijk:
We should implement the Service Info API client.

runcom:
relates to https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/docs/specs/serviceinfo_api.md

PR #210, opened 2022-02-16
feature: add serviceinfo API client

puiterwijk:
This makes the Owner Onboarding Server use the ServiceInfo API for all
ServiceInfo data.

Fixes: #153

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,5 +14,6 @@ members = [
     "manufacturing-client",
     "admin-tool",
 
+    "serviceinfo-api-dev-server",
     "integration-tests",
 ]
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -13,5 +13,6 @@ members = [
     "manufacturing-server",
     "manufacturing-client",
 
+    "serviceinfo-api-dev-server",
```
runcom, line 16: I think I forgot why we call it `-dev-`?

puiterwijk: Because this implementation is incredibly rudimentary (it doesn't allow different serviceinfo for different device), and it's purely here for the integration tests.
For live deployments, people are expected to implement the API themselves.

runcom: right, thanks Patrick!

runcom (approved):

puiterwijk:
@Mergifyio refresh

PR #221, opened 2022-02-23
Ov api

say-paul:
endpoints:
POST /management/v1/ownership_voucher

POST /management/v1/ownership_voucher/delete


owner-onboarding-server/src/ov_reply.rs
```diff
@@ -0,0 +1,85 @@
...
+        if ovs.len() != c {
+                let  e =  Error::generate(ErrorCode::IncompleteVoucher, ovs.len());
+                let e = warp::reply::json(&e);
+                return Ok(warp::reply::with_status(e, StatusCode::BAD_REQUEST));
+        }  
+        
+        
+        
+        // When every check passes
+        let uuid: Vec<String> = ovs.iter()
```
say-paul, line 73: Its not saving ov
@puiterwijk

http-wrapper/src/server.rs
```diff
@@ -483,4 +485,39 @@ where
...
+        .and(warp::path::end())
+        .and(warp::header::exact("Authorization","Bearer"))
+        .and(warp::header::value("X-Number-Of-Vouchers"))
+        .and(warp::body::content_length_limit(1024 * 32))
+        // option 1 works
+        .map(move |c| (c,user_data.clone()))
+        
+        // option 2 doesnt line: 521 complains, and I need to parse the body for the ownership voucher
+        // .and(warp::body::aggregate())
+        // .map(move |c,body| (c,body,user_data.clone()))
```
say-paul, line 517: @puiterwijk This method will put things more in line with the existing framework, but not able to parse the body

owner-onboarding-server/src/ov_reply.rs
```diff
@@ -0,0 +1,85 @@
...
+                .map(|ov| {
+                   let a = ov.clone();
+                    user_data
+                    .ownership_voucher_store
+                    .store_data(a.header().guid().clone(), a);
```
say-paul, line 78: Its not saving ov
@puiterwijk

http-wrapper/src/server.rs
```diff
@@ -483,4 +485,39 @@ where
...
 }
+
+//This method to keep things more inline with the framework
+pub fn ov_request_filter<UDT, F, FR>( 
+    user_data: UDT,
+    handler: F,
+) -> warp::filters::BoxedFilter<(warp::reply::Response,)>
+where 
+    UDT: Clone + Send + Sync + 'static,
+    //option 1 works
```
say-paul, line 498: need some eyes

owner-onboarding-server/src/ov_reply.rs
puiterwijk: Right, because this is a future.
You will need to make sure to await its results, and the compiler should be warning you about this, I believe?

http-wrapper/src/server.rs
puiterwijk: What is the error you get if you use this case?
I'm going to guess it's related to the `&[u8]` lifetime, or the fact that that's not an `impl Buf`?

say-paul: Geez!! the reply notification came to my gmail, which I didn't notice, until today..!!!!!!

data-formats/src/constants.rs
```diff
@@ -1,3 +1,5 @@
+
```
puiterwijk, line 1: Please remove these whitespace changes.

data-formats/src/messages/v11/ov.rs
```diff
@@ -0,0 +1,14 @@
+//not implemented yet
```
puiterwijk, line 1: This is not an FDO v1.1 message, which is the goal for the `data-formats/src/messages/v11` module.

data-formats/src/ownershipvoucher.rs
```diff
@@ -195,6 +195,14 @@ impl OwnershipVoucher {
             .collect()
     }
 
+    //test to get error fro partial parse
+    pub fn pem_read_error_count(data: &[u8]) {
```
puiterwijk, line 199: Why is this in a `pub fn`?

data-formats/src/ownershipvoucher.rs
```diff
@@ -318,6 +326,10 @@ impl OwnershipVoucher {
             .get(OwnershipVoucherIndex::Header as usize)
             .unwrap()
     }
+
+    pub fn contents(&self) -> &ParsedArray<ParsedArraySize5> {
```
puiterwijk, line 330: This is an internal detail that was very explicitly do not want public.
It has happened in the past that the OV format changes (size4 to size5 for example), and the whole purpose of this file is to abstract that away for other code.

owner-onboarding-server/Cargo.toml
```diff
@@ -25,3 +25,4 @@ fdo-data-formats = { path = "../data-formats", version = "0.4.0" }
 fdo-http-wrapper = { path = "../http-wrapper", version = "0.4.0", features = ["server", "client"] }
 fdo-store = { path = "../store", version = "0.4.0", features = ["directory"] }
 fdo-util = { path = "../util", version = "0.4.0" }
+
```
puiterwijk, line 28: Why this whitespace-only change?

owner-onboarding-server/src/handlers.rs
```diff
@@ -655,4 +657,4 @@ pub(crate) async fn report_to_rendezvous_handler(
         .await
         .map_err(|e| warp::reject::custom(RtrFailure(e)))?;
     Ok(warp::reply::Response::new("ok".into()))
-}
+}
```
puiterwijk, line 660: More whitespace changes, and removal of end-of-file-newline.

owner-onboarding-server/src/main.rs
```diff
@@ -394,7 +399,7 @@ async fn main() -> Result<()> {
         // Owner addresses
         owner_addresses,
     });
-
+    let store_path = settings.ownership_voucher_store_config.clone().unwrap().to_string();
```
puiterwijk, line 402: This happens to work with the `Directory` store (also, not after #245), but this is an implementation detail.
You should not ever bypass the Store layer.

owner-onboarding-server/src/ov_management/ov_reply.rs
```diff
@@ -0,0 +1,178 @@
...
+    body: impl Buf,
+    store_path: String,
+    owner_key: PKey<Private>,
+) -> Result<impl warp::Reply,warp::Rejection> 
+{
+    // check for ov type
+    let content_type = content_type.to_str().unwrap();
+    let ovs;
+    match content_type {
+        "application/x-pem-file" => ovs= OwnershipVoucher::many_from_pem(body.chunk()),
```
puiterwijk, line 66: This doesn't implement the `application/cbor` type.

puiterwijk (changes requested):
The actual OV management implementation totally bypasses the Store module, please do not do that.

Additionally, you have a couple of whitespace differences: please clear that up.

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+        let  e =  ErrorMsg::generate(ErrorCode::InvalidNumberOfVouchers, ovs.len());
+        return Ok(warp::reply::with_status(e, StatusCode::BAD_REQUEST));
+    }
+    let mut owned_guid = Vec::new();
+    let mut unowned_guid = Vec::new();
+    for ov in ovs.iter() {
+        if validate_ov(ov,&owner_key){
+            let o = ov.header().guid().to_string();
+            owned_guid.push(o);
+            async {user_data
```
puiterwijk, line 139: This means that if there are some unowned vouchers in the set, the others are actually imported.

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,219 @@
+use std::sync::Arc;
+use openssl::pkey::{Private, PKey};
+use warp::{Filter};
+use std::{usize,str::FromStr};
+use fdo_data_formats::{ownershipvoucher::OwnershipVoucher, types::Guid};
+use serde::{Serialize, Deserialize};
+use warp::{hyper::StatusCode, http::HeaderValue, Buf, reply::Json};
+use crate::OwnerServiceUD;
+
+pub(crate) fn ov_filter(user_data:Arc<OwnerServiceUD>,owner_key: PKey<Private>)
```
puiterwijk, line 10: Why are you passing in the private key here, instead of the public key corresponding to it?
Passing the private key around to code that doesn't need it is an unnecessary risk.

data-formats/src/constants.rs
```diff
@@ -382,7 +382,7 @@ pub enum MessageType {
     DIUNProvideKeyParameters = 213,
     DIUNProvideKey = 214,
     DIUNDone = 215,
-
+
```
puiterwijk, line 385: There are still some whitespace changes.

owner-onboarding-server/src/main.rs
```diff
@@ -426,7 +428,9 @@ async fn main() -> Result<()> {
                 .or(handler_to2_prove_device)
                 .or(handler_to2_device_service_info_ready)
                 .or(handler_to2_device_service_info)
-                .or(handler_to2_done),
+                .or(handler_to2_done)
+                .or(ov_filter(user_data.clone(),user_data.owner_key.clone()))
```
puiterwijk, line 432: If you're already passing in `user_data`, then why are you also cloning the `owner_key` separately?

owner-onboarding-server/src/ov_management/ov_reply.rs
say-paul: Added the functionality

say-paul (commented):
changed accordingly

data-formats/src/constants.rs
```diff
@@ -382,7 +382,6 @@ pub enum MessageType {
     DIUNProvideKeyParameters = 213,
     DIUNProvideKey = 214,
     DIUNDone = 215,
-
```
puiterwijk, line 385: Still whitespace changes remaining.

owner-onboarding-server/src/main.rs
```diff
@@ -34,6 +34,10 @@ use fdo_util::servers::{
 };
 
 mod handlers;
+mod ov_management;
+use crate::ov_management::ov_filter;
+#[macro_use]
+extern crate lazy_static;
```
puiterwijk, line 40: We use rust edition 2015 or later, so we do not need `extern crate` statements.

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    // check-empty
+    let ovs = ovs.unwrap();
+    if ovs.is_empty() {
+        let  e =  ErrorMsg::generate(ErrorCode::ParseError, "Ownership voucher not found!!");
+        return Ok(warp::reply::with_status(e, StatusCode::BAD_REQUEST));
+    }
+
+    // check if X-Number-Of-Vouchers present
+    let count = count.to_str().unwrap().parse::<usize>();
```
puiterwijk, line 75: We shouldn't use `.unwrap()`.
Maybe do something like:
`let count = count.to_str().map_err(|_| ErrorMsg::generate(...)?.parse::<usize>().map_err(|_| ...)?;`

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+
+}
+
+async fn delete_ov(
+    body: impl Buf,
+    user_data: Arc<OwnerServiceUD>,
+) -> Result<Box<dyn warp::reply::Reply>,warp::Rejection> {
+
+    let u = String::from_utf8(body.chunk().to_vec()).expect("Found invalid UTF-8");
+    let uuids_to_delete:Vec<String> = u
```
puiterwijk, line 121: This looks like manual json parsing. That is a very bad idea.
Please just use the warp json features for parsing them, and then also use that to parse to `Guid`.

puiterwijk (changes requested):

owner-onboarding-server/src/ov_management/mod.rs
say-paul: I get that unwrap() will cause error incase of any error, but map_err(|_| warp::reject()) is returning ghost error with status code 500, so handled all the errors before unwrapping , which works as expected now.

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+        m.insert(ErrorCode::NoOwnershipVoucherFound , "Required");
+        m.insert(ErrorCode::UnknownDevice , "unknown");
+        m
+    };
+}
+
+#[derive(Serialize,Deserialize)]
+struct ErrorMsg<T> {
+    error_code: Option<ErrorCode>,
+    error_details : HashMap<String,T>
```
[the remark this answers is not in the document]
say-paul: Not sure why the use of generic is a problem here, it provides the flexibility of all the required object type  that the response requires, and the overall response is converted to Json before sending.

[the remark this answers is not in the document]
say-paul: Changed it now,

owner-onboarding-server/src/ov_management/mod.rs
say-paul: created a a new function to handle this

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+
+
+const ACCEPTED_CONTENT_TYPE:[&str; 2] = ["application/x-pem-file","application/cbor"];
+
+#[derive(Clone)]
+#[derive(Serialize,Deserialize)]
+#[derive(Eq,PartialEq,Hash)]
+#[derive(Error,Debug)]
+enum ErrorCode {
+    #[error("error")]
```
puiterwijk, line 174: The argument to the `#[error()]` call should be an actual error description.
For example: `#[error("Parse error: {0}")]`.
It would also be great to actually pass the error message along with this perhaps, like `ParseError(crate::Error)`, so that that information is passed along via this.

Doing both of those, you could get mostly rid of the `generate` and `generate_with_Description` methods that take extra arguments, and instead just have the `ErrorCode` struct generate those from the value.

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+    IncompleteVoucher,
+    #[error("{0}")]
+    PemParseErrorDescription(String),
+    #[error("parsed correctly")]
+    PemParseError,
+}
+
+#[derive(Serialize)]
+struct ErrorMsg {
+    error_code: Option<ErrorCode>,
```
puiterwijk, line 204: When would `error_code` ever be `None`?

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+
+}
+
+fn validate_ov<'a> (ov: &'a OwnershipVoucher,
+    owner_key: &'a PKey<Private>,
+) -> Result<bool,ErrorCode> {
+        ov.iter_entries()
+        .map_err(|e| ErrorCode::PemParseErrorDescription(e.to_string()))?
+        .into_iter().next()
+        .unwrap()
```
puiterwijk, line 244: I think this would panic if it's an OV without any entries?

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+        .map_err(|e| ErrorCode::IncompleteVoucherDescription(e.to_string()))?
+        .public_key()
+        .matches_pkey(owner_key)
+        .map_err(|e| ErrorCode::InvalidVoucherSignatureDescription(e.to_string()))
```
puiterwijk, line 248: Why do you store the `String` version of the `Error` in the `InvalidVoucherSignatureDescription` key?
You could store the full `Error` at this point, and then only turn it to a `String` when actually serializing.

owner-onboarding-server/src/ov_management/mod.rs
say-paul: Its handled in line:75

puiterwijk: No, that is no Ownership Vouchers. This comment was about an Ownership Voucher that had no ownership transfer entries.

owner-onboarding-server/src/ov_management/mod.rs
say-paul: tried to modify errorCodes to:
PemParseErrorDescription(fdo_data_formats::Error)

but getting error : trait bound `fdo_data_formats::Error: _::_serde::Serialize` is not satisfied
the trait `_::_serde::Serialize` is not implemented for `fdo_data_formats::Error`

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+use warp::{hyper::StatusCode, http::HeaderValue, Buf};
+use crate::{OwnerServiceUD};
+use thiserror::Error;
+use serde_json::{json};
+
+
+pub(crate) fn ov_filter(session_store:Arc<OwnerServiceUD>) 
+-> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone
+{       
+        let store1 = session_store.clone();
```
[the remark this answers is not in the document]
say-paul: on dropping this and adding session_store.store to both of the filters,

getting error: move occurs because `session_store` has type `std::sync::Arc<OwnerServiceUD>`, which does not implement the `Copy` trait

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+        "application/x-pem-file" => OwnershipVoucher::many_from_pem(body.chunk()),
+        "application/cbor" => OwnershipVoucher::deserialize_many_from_reader(body.chunk()),
+        _ => {
+            return ErrorMsg::generate(ErrorCode::RequireContentTypeHeader, ACCEPTED_CONTENT_TYPE);
+        }
+    };
+
+    // parse-error
+    if ovs.is_err() {
+        return ErrorMsg::generate(ErrorCode::ParseError, ovs.unwrap_err().to_string());
```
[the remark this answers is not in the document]
say-paul: changed the implementation so may not be required anyway

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+            }
+        }
+    }
+    
+    if !unowned_guid.is_empty() {
+        return ErrorMsg::generate(ErrorCode::UnwonedVoucher, unowned_guid);
+    }
+    
+    for ov in ovs.iter() {
+            let _ = user_data
```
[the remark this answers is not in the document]
say-paul: added.

owner-onboarding-server/src/ov_management/mod.rs
say-paul: added

owner-onboarding-server/src/ov_management/mod.rs
say-paul: Updated

owner-onboarding-server/src/ov_management/mod.rs
say-paul: dropped this Error as we are reporting custom error message with no much of details

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+        .and(warp::post())
+        .and(warp::path("ownership_voucher"))
+        .and(warp::path("delete"))
+        .and(warp::path::end())
+        .and(warp::header::exact("Content-Type", "Application/json"))
+        .and(warp::body::content_length_limit(1024 * 32))
+        .and(warp::body::json())
+        .map(move |uuids: Vec<String>| (uuids, store1.clone()))
+        .untuple_one()
+        .and_then(delete_ov);
```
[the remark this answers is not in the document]
say-paul: added recover

PR #222, opened 2022-02-23
feat(client): send errors back while performing TO1/TO2

7flying:
Implements #9, sends back errors to RV while performing TO1 and TO2.
* Changes related to handling errors on TO2.
* Changes related to handling errors on TO1.


puiterwijk:
Just a general feeling: the overall code looks like it would work, but I guess my main concern right now is that this is a lot of repeated code.
Maybe it's an idea to break the different bits from different message pairs up, i.e. an `fn` for the `HelloRV -> HelloRVAck` and related processing, and then have the handling of errors happen at that level?
i.e. `fn perform_hellorv() -> Result<...>`, and then:
```
let (NonceTO1Proof , eBSigInfo ) = match perform_hellorv() {
    Err(e) => send_error(...),
    Ok(res) => res,
};
let ... = matchperform_provetorv(NonceTO1Proof, eBSigInfo) {
.....
```

7flying:
Yes, it is very repetitive. I will break it up in functions that handle different message pairs, as you say :+1:

client-linuxapp/src/main.rs
```diff
@@ -104,57 +156,154 @@ async fn get_client_list(rv_entry: &RendezvousInterpretedDirective) -> Result<Ve
...
         devcred.device_guid().clone(),
         SigInfo::new(sig_type, vec![]),
     );
     let hello_rv_ack: RequestResult<messages::v11::to1::HelloRVAck> =
         client.send_request(hello_rv, None).await;
-    let hello_rv_ack = hello_rv_ack.context("Error sending HelloRV")?;
+
+    let hello_rv_ack = match hello_rv_ack {
+        Ok(hello_rv_ack) => hello_rv_ack,
+        Err(e) => {
```
puiterwijk, line 175: Instead of this pattern match all the time, you could do:
```
let hello_rv_ack = hello_rv_ack
    .context("Error sending HelloRV")
    .map_err(|e| ClientError::Request(ErrorResult::new(
      ErrorCode::InternalServerError,
      String::from("Error sending HelloRV"),
      MessageType::TO1HelloRV,
      e,
))?;
```

This basically transforms (with `.context()`) from a `Result<_, fdo_http_wrapper::client::Error>` to `Result<_, anyhow::Error>`, and then (with `.map_err()`) to `Result<_, ClientError>`, and then with the `?` operator, it uses the normal "return if it errored" mechanism.

client-linuxapp/src/main.rs
```diff
@@ -104,57 +156,154 @@ async fn get_client_list(rv_entry: &RendezvousInterpretedDirective) -> Result<Ve
...
+            )))
+        }
+    };
     log::trace!("Hello RV ack: {:?}", hello_rv_ack);
 
-    // Check ack
     let b_sig_info = hello_rv_ack.b_signature_info();
     if b_sig_info.sig_type() != sig_type {
-        bail!("Unsupported sig type returned");
+        return Err(ClientError::Response(ErrorResult::new(
```
puiterwijk, line 188: In these cases (where there is no other error), this form is good.

client-linuxapp/src/main.rs
```diff
@@ -38,6 +43,53 @@ const RV_DEFAULT_DELAY_SEC: f32 = 120.0;
 const RV_DEFAULT_DELAY_OFFSET: f32 = 30.0;
 const RV_USER_DEFINED_DELAY_OFFSET: f32 = 0.25;
 
+// Encapsulates errors caused during TO1/TO2
+#[derive(Debug)]
+struct ErrorResult {
+    e_code: ErrorCode,
+    e_string: String,
```
puiterwijk, line 50: As far as I can find, this is always just a hardcoded string that you type in the code explicitly.
You could use `e_string: &'static str` instead then, that avoids having to create a `String` object, and is slightly neater code throughout.
Normally, when you have a borrowed object in a `struct`, it needs an explicit lifetime that the `struct` itself also has, but `'static` is special, in that it doesn't require that: that indicates that the (in this case string) is statically compiled into the binary, so it has a lifetime that exceeds any function we define.

puiterwijk (commented):
Thanks for the rework.
A few more remarks/requests for change, but we're getting close!

puiterwijk (approved):
Thanks for this patch!

puiterwijk:
@Mergifyio refresh

Issue #200, opened 2022-02-08
Device Credential: semi-TPM


puiterwijk:
We can do a Device Credential system where the DC is in a filesystem, but it doesn't contain the private key (or HMAC key), but just the data used to ask the TPM to reproduce the key.

mangelajo:
Yeah, that's how I found the tpm2 interface works mostly, it will provide you with a blob that can be passed back to the TPM module to give you the key. In a way that the tpm doesn't really need to store all the blobs for all generated keys, it just generates, encrypts, gives it to you, and you save it.

At least that's my overall understanding, I could be wrong.

puiterwijk:
You're right with that, but our initial intent was to store everything in the TPM, including the rendezvous info and the data to derive the keys, so that there's nothing in the filesystem that could be accidentally erased before onboarding.
This ticket is to get a hybrid approach, awaiting some standardization processes.

mangelajo:
All this also makes me think, off-chip TPMs are probably insecure, unless I'm missing some detail, for example, can't the secrets be dumped by just listening to the SPI bus? , I hope there's some additional crypto involved in those communications.

Of course, tpm "stored" PKs that will perform signatures/encryption in-tpm shouldn't be a big problem, as the final form of those is never leave the chip. The problem I see is if those keys leave the TPM to be used by the host...

nullr0ute:
TPMs have the ability to encrypt traffic over the transport (SPI, I2C etc) if it's used properly so as such while they can't be insecure it doesn't mean they actually are if used properly.

puiterwijk:
The keys I intended are the on-chip kind that never leave the TPM.
The TPM would just return the public key, and then during onboarding, it would regenerate the same key and sign things.

PR #224, opened 2022-02-24
feat: implement semi-TPM Device Credential

puiterwijk:
This implements a system where the keys for the device credential are
created by the TPM (never leaving the TPM), but the references to the
keys, and rendezvous info, etc, are stored on the filesystem.
This protects against having the actual private keys stolen when the
device credential file is extracted from a disk.

The current implementation only supports the ECDSA ciphers mentioned by
the FDO specification.

The current test suite does not set up a software TPM to execute the TPM
tests during CI, but it will execute the TPM tests if a TPM exists on
the system.

Fixes: #200

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,7 @@ jobs:
   clippy:
     name: Clippy
     runs-on: ubuntu-latest
+    container: fedora:latest
     steps:
       - uses: actions/checkout@v2
       - name: Cache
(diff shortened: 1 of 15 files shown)


puiterwijk:
This is still missing tests.

runcom:
code lgtm - I guess the tricky part would just be to update the tests in composer to run vms with `--tpm backend.type=emulator,backend.version=2.0,model=tpm-crb`

puiterwijk:
@runcom do note that the File-backed DC's are still possible. And the server can choose which types are acceptable.

runcom (approved):

mangelajo:
Great work 👌🏻

PR #238, opened 2022-03-01
chore: update sample owner config file

7flying:
The sample config files do not work as they are without the new required fields (`service_info_api_url` and `service_info_api_authentication`), so incorporate them in the sample config.


examples/config/owner-onboarding-server.yml
```diff
@@ -13,6 +13,10 @@ owner_addresses:
     - dns_name: fdo.example.com
 report_to_rendezvous_endpoint_enabled: false
 bind: 0.0.0.0:8081
+service_info_api_url: "http://localhost:8089/device_info"
+  service_info_api_authentication:
```
runcom, line 17: indentation is off here @7flying - this should be top level

7flying: I added that using the integration test [example](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/integration-tests/templates/owner-onboarding-server.yml.j2). Which one do you mean that should be top level?

examples/config/owner-onboarding-server.yml
```diff
@@ -13,6 +13,10 @@ owner_addresses:
     - dns_name: fdo.example.com
 report_to_rendezvous_endpoint_enabled: false
 bind: 0.0.0.0:8081
+service_info_api_url: "http://localhost:8089/device_info"
+service_info_api_authentication:
+  BearerToken:
+    token: TestAuthToken
 service_info:
```
puiterwijk, line 20: This entire bit should now be removed.

examples/config/owner-onboarding-server.yml
```diff
@@ -13,27 +13,7 @@ owner_addresses:
...
-    return_stdout: true
-    return_stderr: true
-  - command: touch
-    args:
-    - /tmp/command-testfile
\ No newline at end of file
+service_info_api_url: "http://localhost:8089/device_info"
+service_info_api_authentication:
+  BearerToken:
+    token: TestAuthToken
```
runcom, line 19: this example should just be:
```toml
+service_info_api_url: "file:///path/to/service_info.json"
+service_info_api_authentication: None
```
as this is what we ship in RPMs too and since we don't ship a service info server we need this

runcom: ok file:// doesn't work also

examples/config/owner-onboarding-server.yml
```diff
@@ -13,27 +13,5 @@ owner_addresses:
...
-    args:
-    - /etc/doesnotexist/test
-    may_fail: true
-    return_stdout: true
-    return_stderr: true
-  - command: touch
-    args:
-    - /tmp/command-testfile
\ No newline at end of file
+service_info_api_url: "file:///path/to/service_info.json"
```
runcom, line 16: ok file:// doesn't work - let's hold this, we need something that an user can play with and we don't currently package any service info server

runcom:
@7flying can you rebase to main? I'll merge afterwards

7flying:
Closing 'cause I'm unable to clean history. See #248

PR #267, opened 2022-05-12
chore(examples): clarify paths again

runcom:
a similar change was dropped when the sim server was split from the
onboarding server. Put this back.


runcom:
@puiterwijk seems a  git diff error too? :thinking:

puiterwijk:
@runcom fixed in #271. Please rebase, and then it should work.

PR #277, opened 2022-06-06
build: update spec file to add man pages

### LINKED ISSUE #269: FDO project doesn't ship with documentation
We should have man pages to ship them in RPMs too (this comes from a user's feedback, go online isn't an option)

7flying:
Update rpm build file to generate man pages from rst and add those to the packages.

Fixes #269


puiterwijk (commented):
Cool, thanks!
Maybe it's an idea to run the build during CI at some stage or another?

puiterwijk (approved):
Awesome, thanks!

Issue #289, opened 2022-07-07
Create the target directory if it does not pre-exist on the system


luisarizmendi:
FDO fails when trying to copy files to a destination path that does not pre-exist on the target device. 

FDO client should create the required directory.

For example, if you try to create the hello.txt file in ‘/var/home/admin/new/’ (which does not exist in the base image)

[serviceinfo_api_server.yml]

```
…
  - path: /var/home/admin/new/hello.txt
    permissions: 644
    source_path: /etc/fdo-configs/device0/var/home/admin/new/hello.txt
…
```

This is the error that you will find:

```
Jul 07 03:43:31 localhost.localdomain systemd[1]: Starting FDO client...
Jul 07 03:43:31 localhost.localdomain fdo-client-linuxapp[1251]:  2022-07-07T07:43:31.050Z INFO  fdo_client_linuxapp > Found device credential at FileSystemPath { path: "/boot/device-credentials", deactivation_method: None }
Jul 07 03:43:31 localhost.localdomain fdo-client-linuxapp[1251]:  2022-07-07T07:43:31.109Z INFO  fdo_client_linuxapp > Got TO2 addresses: ["http://192.168.122.200:8081", "http://fe80::5054:ff:fe3b:326d:8081"]
Jul 07 03:43:31 localhost.localdomain fdo-client-linuxapp[1251]:  2022-07-07T07:43:31.109Z INFO  fdo_client_linuxapp > Performing TO2 protocol, URL: "http://192.168.122.200:8081"
Jul 07 03:43:31 localhost.localdomain fdo-client-linuxapp[1251]:  2022-07-07T07:43:31.174Z INFO  fdo_client_linuxapp::serviceinfo > Creating file "/var/home/admin/new/hello.txt" with 5 bytes (mode 420)
Jul 07 03:43:31 localhost.localdomain fdo-client-linuxapp[1251]:  2022-07-07T07:43:31.174Z INFO  fdo_client_linuxapp              > ServiceInfo failed, error: Error processing returned serviceinfo
Jul 07 03:43:31 localhost.localdomain fdo-client-linuxapp[1251]: Caused by:
Jul 07 03:43:31 localhost.localdomain fdo-client-linuxapp[1251]:     0: Error deploying binary file
Jul 07 03:43:31 localhost.localdomain fdo-client-linuxapp[1251]:     1: Error creating file
[...]

PR #292, opened 2022-07-07
fix(client-linuxapp): refactor binary file in progress methods

runcom:
Also add some unit testing. What this does is basically:

- make sure the path we get is always absolute, and do not join it with
  the test variable before that check
- make sure paths are joined correctly in case there's a prefix from
  environment variables (mainly for testing)
- refactor the struct/methods to be easy to add some unit test to
  validate what we're writing
- also make sure we create the parents (mkdir -p) when writing a file (Fixes https://github.com/fedora-iot/fido-device-onboard-rs/issues/289)

### DIFF
--- a/client-linuxapp/src/serviceinfo.rs
+++ b/client-linuxapp/src/serviceinfo.rs
@@ -129,18 +129,20 @@ fn perform_rhsm(organization_id: &str, activation_key: &str, perform_insights: b
 }
 
 #[derive(Debug)]
-struct BinaryFileInProgress {
+struct BinaryFileInProgress<'a> {
     path: Option<String>,
+    prefix: Option<&'a str>,
     length: Option<u64>,
     contents: Option<Vec<u8>>,
     mode: Option<u32>,
     digest: Option<Hash>,
 }
 
-impl BinaryFileInProgress {
-    fn new() -> Self {
+impl<'a> BinaryFileInProgress<'a> {
+    fn new(prefix: Option<&'a str>) -> Self {
         BinaryFileInProgress {
             path: None,
+            prefix,
             length: None,
             contents: None,
             mode: None,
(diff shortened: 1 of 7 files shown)


client-linuxapp/src/serviceinfo.rs
```diff
@@ -131,36 +131,44 @@ fn perform_rhsm(organization_id: &str, activation_key: &str, perform_insights: b
...
+    fn destination_path(path_str: &String, prefix: Option<String>) -> Result<PathBuf> {
+        let path = PathBuf::from(path_str);
 
         if !path.is_absolute() {
             bail!("Binary file path must be absolute");
         }
 
+        if let Some(val) = prefix {
+            // make sure we drop the first slash or PathBuf returns the join'ed absolute path
+            Ok(PathBuf::from(&val).join(&path_str[1..path_str.len()]))
```
puiterwijk, line 162: Fun thing: this assumes POSIX, i.e. non-Windows.

client-linuxapp/src/serviceinfo.rs
```diff
@@ -358,7 +368,8 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
     let mut rhsm_activation_key: Option<String> = None;
     let mut rhsm_perform_insights: Option<bool> = None;
 
-    let mut binary_file_in_progress = BinaryFileInProgress::new();
+    let binary_file_prefix = env::var("BINARYFILE_PATH_PREFIX")?;
+    let mut binary_file_in_progress = BinaryFileInProgress::new(Some(binary_file_prefix.clone()));
```
puiterwijk, line 372: Would it be possible to pass it in as a `&str` instead, so we don't have to clone all the time?
Also, this seems to imply that `binary_file_prefix` is never empty?
Maybe `let binary_file_prefix = env::var("...").ok()`, to make it an `Option`, and then pass it in as `Option<&str>`, or even `Option<Path>`?

client-linuxapp/src/serviceinfo.rs
runcom: don't we assume client-linuxapp only runs on linux which should mostly be posix?

puiterwijk: I have tried to ensure we don't always assume that. I don't see a reason why it couldn't run on Windows.

client-linuxapp/src/serviceinfo.rs
runcom: I'll check for emptiness - as for having a &str instead of a String it's messy afaict but I'll see what I can come up with (lifetime and Some(&str) being messy too etc etc)

runcom: fixed

client-linuxapp/src/serviceinfo.rs
runcom: fixed, as per your suggestion, I've added a conditional compilation and error out on !unix

runcom:
all good :)

puiterwijk (approved):
Cool, thanks!

PR #293, opened 2022-07-07
bump config crate to 0.12

nullr0ute:
Minor bump of the config crate to a newer rev.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>


nullr0ute:
This will make deps a little easier to handle in Fedora for the time being, it's not bumping to the latest 0,13.x as that makes it worse :-/

Issue #132, opened 2021-11-10
Implement OV Management API


puiterwijk:
After #107 gets merged, we should implement the Ownership Voucher management API in the owner onboarding server.

PR #299, opened 2022-07-26
feat(ov_api): add and delete ownership voucher in onboarding server

say-paul:
Closes: #132 
endpoints:
POST /management/v1/ownership_voucher

POST /management/v1/ownership_voucher/delete


nullr0ute:
This needs at least a decent descriptive commit message, a header with "working model" isn't really enough, What does the patch do? From the commit message I can not tell. Also possibly a sign off line (Not sure if we explicitly require them).

say-paul:
Added test and fixed the commit message

nullr0ute:
Closing this for the time being as it needs work, we'll revisit when we have a more defined requirement.

Issue #268, opened 2022-05-16
fdo client and serviceinfo api server doesn't log which command failed in a list of commands


runcom:
As per the title, not fully sure about the serviceinfo api server, but it's pretty critical to provide logs in the client perhaps

7flying:
Closing as done as this was already fixed by #307

PR #301, opened 2022-07-28
fix: fdo client and serviceinfo api server doesn't log which command failed in a list of commands #268

sarmahaj:
This pull request contains a small change to fix above bug.
Also in file client-linuxapp/src/serviceinfo.rs , inside fn execute(self, si_out: &mut ServiceInfo) -> Result<()> , I had to comment line 328-335 , since this line of code was ignoring stderr and I could not print stderr with it.


client-linuxapp/src/serviceinfo.rs
```diff
@@ -323,12 +324,15 @@ impl CommandInProgress {
         let mut cmd = Command::new(self.command.as_ref().unwrap());
         cmd.args(&self.args);
 
-        if !self.return_stdout {
+        /*Below code ignores stderr incase of failure, commenting this part*/
+        /*
+         if !self.return_stdout {
```
7flying, line 329: If this is part of the bug you can safely delete it instead of commenting it

client-linuxapp/src/serviceinfo.rs
```diff
@@ -19,6 +19,7 @@ use fdo_data_formats::{
     types::{CborSimpleTypeExt, Hash, ServiceInfo},
 };
 use fdo_http_wrapper::client::{RequestResult, ServiceClient};
+//use tokio::io::stderr;
```
7flying, line 22: We can clean this comment

client-linuxapp/src/serviceinfo.rs
sarmahaj: done now ,thanks @7flying

nullr0ute:
I think we can squash all 3 commits into a single commit and please add a better commit log. Do a short description on the first line and then a larger description underneath.

sarmahaj:
> I think we can squash all 3 commits into a single commit and please add a better commit log. Do a short description on the first line and then a larger description underneath.

done , squashed the commits in one commit.

client-linuxapp/src/serviceinfo.rs
```diff
@@ -20,6 +20,7 @@ use fdo_data_formats::{
 };
 use fdo_http_wrapper::client::{RequestResult, ServiceClient};
 
+
```
nullr0ute, line 23: Probably don't need a new line here

nullr0ute (commented):
The commit should look something like:

````
fix(client-linuxapp): Fix logging for serviceinfo api server

In the client serviceinfo api server doesn't log which command failed
in a list of commands which makes it hard for users to determine what
is failing and how to deal with it. Fix up the logging to make it easier for
users to see what's going on when things don't work. 

Signed-off-by: My Name <email@xxx.com>
````

There's other good examples of commit messages in the git repo to use as examples and you can find details of sign offs on the internet.

File: client-linuxapp/src/serviceinfo.rs [hunk 4]
```diff
@@ -323,13 +323,6 @@ impl CommandInProgress {
         let mut cmd = Command::new(self.command.as_ref().unwrap());
         cmd.args(&self.args);
 
-        if !self.return_stdout {
-            cmd.stdout(std::process::Stdio::null());
-        }
-        if !self.return_stderr {
-            cmd.stderr(std::process::Stdio::null());
-        }
-
```
runcom, line 332: @puiterwijk ptal

nullr0ute:
Can you do a git rebase and git push so we don't have a bunch of merge requests in the commit.

7flying:
It seems that one of those merge commits removes the files from the PR and we don't get any changes. I would suggest safely stashing the commit with the changes, not the merges, resetting the branch, getting the latest changes from main, applying the commit and then force-pushing everything.

7flying:
Closing as the bug was already fixed by #307

PR #302, opened 2022-07-28
fix(inital user creation) : create user during onboarding

### LINKED ISSUE #270: initial_user doesn't create the user
reported by a user - the user has to exist - and this seems to be a bug as the intention should be to create the user (haven't tested myself, just reported, let's validate and fix if needed)

say-paul:
Fixes #270 

This shall create initial user during the onboarding process.

### DIFF
--- a/client-linuxapp/src/serviceinfo.rs
+++ b/client-linuxapp/src/serviceinfo.rs
@@ -52,6 +52,23 @@ fn set_perm_mode(path: &Path, mode: u32) -> Result<()> {
     Ok(())
 }
 
+fn create_user(user: &str) -> Result<()> {
+    let user_info = passwd::Passwd::from_name(user);
+    if user_info.is_some() {
+        log::info!("User: {} already present", user);
+        return Ok(());
+    }
+    log::info!("Creating user: {}", user);
+    Command::new("useradd")
+        .arg("-m")
+        .arg(user)
+        .spawn()
+        .context("Error spawning new user command")?
+        .wait()
+        .context("Error creating new user")?;
+    Ok(())
+}
+
 fn install_ssh_key(user: &str, key: &str) -> Result<()> {
     let user_info = passwd::Passwd::from_name(user);
     if user_info.is_none() {
@@ -604,6 +621,10 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
         if sshkey_user.is_none() || sshkey_key.is_none() {
             bail!("SSHkey module missing username or key");
         }
+        create_user(sshkey_user.as_ref().unwrap()).context(format!(
+            "Error creating new user: {}",
+            sshkey_user.as_ref().unwrap()
+        ))?;
         install_ssh_key(sshkey_user.as_ref().unwrap(), sshkey_key.as_ref().unwrap())
             .context("Error installing SSH key")?;
     }
--- a/integration-tests/tests/e2e.rs
+++ b/integration-tests/tests/e2e.rs
@@ -1,4 +1,5 @@
 mod common;
+use std::env;
 #[allow(unused_imports)]
 use std::{fs, io::Write, process::Command, time::Duration};
 
@@ -108,10 +109,10 @@ async fn test_e2e_impl<F>(verification_generator: F, diun_key_type: &str) -> Res
 where
     F: Fn(&TestContext) -> Result<(&'static str, String, &'static str)>,
 {
+    let ci = env::var("FDO_PRIVILEGED").is_ok();
     let mut ctx = TestContext::new().context("Error building test context")?;
-
+    let new_user: &str = "testuser"; //new user to be created during onboarding
     let encrypted_disk_loc = ctx.testpath().join("encrypted.img");
-
     let rendezvous_server = ctx
         .start_test_server(
             Binary::RendezvousServer,
@@ -128,6 +129,9 @@ where
                         "encrypted_disk_label",
                         &encrypted_disk_loc.to_string_lossy(),
                     );
+                    if ci {
+                        cfg.insert("user", new_user)
+                    };
                     Ok(())
                 })?)
             },
@@ -387,6 +391,18 @@ testkey
 # End of FIDO Device Onboarding keys
 "
     );
+    if ci {
+        L.l("Running create initial user validation");
+        pretty_assertions::assert_eq!(
+            passwd::Passwd::from_name(new_user).is_some(),
+            true,
+            "User: {} is not created during onboarding",
+            &new_user
+        );
+    } else {
+        L.l("Skipped create initial user validation
+        To validate set env variable FDO_PRIVILEGED and run test as superuser");
+    }
 
     L.l("Checking encrypted disk image");
     let output = Command::new("cryptsetup")
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,8 @@ jobs:
           command: build
       - name: Run tests
         uses: actions-rs/cargo@v1
+        env:
+          FDO_PRIVILEGED: true
         with:
           command: test
       # This is primarily to ensure that changes to fdo_data.h are committed,
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -31,6 +31,7 @@ serde_cbor = "0.11"
 serde_json = "1.0"
 pretty_assertions = "1.0.0"
 paste = "1.0"
+passwd = "0.0.1"
 pem = "1.0"
 users = "0.11.0"
 


client-linuxapp/src/serviceinfo.rs
```diff
@@ -52,6 +52,22 @@ fn set_perm_mode(path: &Path, mode: u32) -> Result<()> {
     Ok(())
 }
 
+fn create_user(user: &str) -> Result<()> {
+    let user_info = passwd::Passwd::from_name(user);
+    if user_info.is_none() {
+        log::info!("Createing user: {}", user);
+        Command::new("useradd")
+            .arg(format!("-m {}", user))
```
puiterwijk, line 60: This would lead to a single command "-m {username}", so it would end up getting called like: `useradd "-m someusername"`, which isn't the way things are expected.
You should do a `.arg("-m").arg(user)`, to make sure it becomes to separate arguments.

nullr0ute:
A few general pieces:
1) more detailed commit message
2) Signed off line
3) Tests

client-linuxapp/src/serviceinfo.rs
```diff
@@ -52,6 +52,23 @@ fn set_perm_mode(path: &Path, mode: u32) -> Result<()> {
...
+            .arg(user)
+            .spawn()
+            .context("Error spawning new user")?
+            .wait()
+            .context("Error Creating new user")?;
+    } else {
+        log::info!("User: {} already present", user);
+    }
+    Ok(())
+}
```
runcom, line 70: just for readability I'll flip the conditions (just a nit also)

```rust
fn create_user(user: &str) -> Result<()> {
    let user_info = passwd::Passwd::from_name(user);
    if user_info.is_some() {
        log::info!("User: {} already present", user);
        return Ok(());
    }
    log::info!("Creating user: {}", user);
    Command::new("useradd")
        .arg("-m")
        .arg(user)
        .spawn()
        .context("Error spawning new user")?
        .wait()
        .context("Error Creating new user")?;
}
```

client-linuxapp/src/serviceinfo.rs
```diff
@@ -52,6 +52,23 @@ fn set_perm_mode(path: &Path, mode: u32) -> Result<()> {
...
 
+fn create_user(user: &str) -> Result<()> {
+    let user_info = passwd::Passwd::from_name(user);
+    if user_info.is_none() {
+        log::info!("Creating user: {}", user);
+        Command::new("useradd")
+            .arg("-m")
+            .arg(user)
+            .spawn()
+            .context("Error spawning new user")?
```
runcom, line 63: this context doesn't really help as this isn't "spawning new user" but the command itself to create a user

client-linuxapp/src/serviceinfo.rs
```diff
@@ -52,6 +52,23 @@ fn set_perm_mode(path: &Path, mode: u32) -> Result<()> {
...
+            .wait()
+            .context("Error Creating new user")?;
```
runcom, line 65: lowercase Creating

client-linuxapp/src/serviceinfo.rs
```diff
@@ -606,6 +623,7 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
         if sshkey_user.is_none() || sshkey_key.is_none() {
             bail!("SSHkey module missing username or key");
         }
+        create_user(sshkey_user.as_ref().unwrap()).context("Error creating user")?;
```
runcom, line 626: there are now 2 contexes with the very same text, so either it adds more information or I'd remove the one within `create_user`

runcom:
apart from a comment, lgtm

cc @puiterwijk

client-linuxapp/src/serviceinfo.rs
```diff
@@ -626,6 +645,8 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
         create_user(sshkey_user.as_ref().unwrap()).context("Error creating user")?;
         install_ssh_key(sshkey_user.as_ref().unwrap(), sshkey_key.as_ref().unwrap())
             .context("Error installing SSH key")?;
+        set_passwordless_login(sshkey_user.as_ref().unwrap())
+            .context("Error setting up passwordless login")?;
```
runcom, line 649: hold on - create another PR for this, this doesn't belong here as here we're just fixing the user creation

runcom:
@say-paul please drop the last commit, create another PR for that and squash the other commits into one

client-linuxapp/src/serviceinfo.rs
say-paul: make sense split into #306

miabbott:
I believe this fixes https://github.com/fedora-iot/fido-device-onboard-rs/issues/270

runcom:
> I believe this fixes #270

it does

say-paul:
@runcom the request changes are made.

nullr0ute:
We need some tests for this as well before it can be merged.

say-paul:
@nullr0ute test added to run the servericeinfo_api_server and the clientlinuxapp binaries with the user details(test process user)  pushed in the config.Any random user creation during onboarding can only be tested manually though.

client-linuxapp/src/serviceinfo.rs
```diff
@@ -606,6 +623,7 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
         if sshkey_user.is_none() || sshkey_key.is_none() {
             bail!("SSHkey module missing username or key");
         }
+        create_user(sshkey_user.as_ref().unwrap())?;
```
runcom, line 626: I'd add a `.context(..)` here to bubble up errors meaningfully

integration-tests/tests/e2e.rs
```diff
@@ -128,6 +128,10 @@ where
                         "encrypted_disk_label",
                         &encrypted_disk_loc.to_string_lossy(),
                     );
+                    cfg.insert(
+                        "user",
+                        users::get_effective_username().unwrap().to_str().unwrap(),
+                    );
```
runcom, line 134: this just add the user, but is it checked anywhere? just making sure that the user __is indeed__ into the system

say-paul: creating user requires privileges, the way the test works is using a user(process owner) which is already present, so if you try with any random user name, the e2e test will fail. its kind of hack-y, so I lean on testing creation of new user manually or some complex automation of setting up the an VM and do the actual onboarding.

nullr0ute: The fdo client app runs as a privileged user so the tests should run as such no? The whole point of CI is test things automatically so that regressions from other PRs to this feature are appropriately picked up and we don't regress.

say-paul: Seems the test does run as superuser , changed the approach of the e2e test to actually do the onboarding instead of considering it as done.

integration-tests/tests/e2e.rs
```diff
@@ -109,9 +110,8 @@ where
     F: Fn(&TestContext) -> Result<(&'static str, String, &'static str)>,
 {
     let mut ctx = TestContext::new().context("Error building test context")?;
-
+    let new_user: &str = "testguy"; //new user to be created during onboarding
```
runcom, line 113: let's call this `testuser`

integration-tests/tests/e2e.rs
```diff
@@ -345,7 +346,7 @@ where
 
     // Execute TO1/TO2 protocols
     let ssh_authorized_keys_path = ctx.testpath().join("authorized_keys");
-    let marker_file_path = ctx.testpath().join("marker");
+    let marker_file_path = PathBuf::from("/etc/device_onboarding_performed"); //setting the marker file default value from clientlinuxapp
```
runcom, line 349: this is gonna use the system, why changing this?

runcom: So if I run the test on my system that file is going to show up?

say-paul: yes, it will, and you have to `rm' the file before running another test for the onboarding process to run.

runcom: I don't understand the difference between the mocked marker file and the default path - they do the very same, what is this change actually doing?

runcom: the problem here is mainly around running privileged tests in environments like local dev - maybe we just need to skip privileged tests @puiterwijk ? 👼

say-paul: @runcom More I think there can be a way move into the the file location to any custom path, but creating user itself require privileges.

runcom: if this test requires privileges we need to wrap it under some condition so it's not run on my local dev environment cause I don't want an user to be run in my local laptop - CI can expose an env variable to say "I'm in CI" and then this test is run only under that condition

say-paul: @runcom  @nullr0ute updated the test accordingly.

integration-tests/tests/e2e.rs
```diff
@@ -108,10 +109,10 @@ async fn test_e2e_impl<F>(verification_generator: F, diun_key_type: &str) -> Res
 where
     F: Fn(&TestContext) -> Result<(&'static str, String, &'static str)>,
 {
+    let ci = env::var("FDO_TEST_MODE").is_ok();
```
runcom, line 112: I'd call this `FDO_PRIVILEGED` since we're not even looking what's inside anyway

runcom:
perfect, can you just squash your commits down to just one and rebase?

runcom (approved):

PR #304, opened 2022-07-29
fix: update spec for multiple vouchers

rdotjain:
For uploading multiple ownership vouchers, the spec says, `vouchers should be appended to each other, adding a newline between the different entries`. But this newline isn't required; the code can split vouchers on encountering a new initial tag (-----BEGIN OWNERSHIP VOUCHER-----)


nullr0ute:
We could also word it like ```adding a newline or separated by begin and end tags between the different entries.```

rdotjain:
Oh yes, that would make more sense. Thanks

rdotjain:
> We could also word it like `adding a newline or separated by begin and end tags between the different entries.`

Oh yes, that would make more sense. Thanks

rdotjain (commented):
updated spec

nullr0ute:
You can squash them into a single commit and we're good to go

nullr0ute (approved):
LGTM

PR #305, opened 2022-08-01
chore(client-linuxapp): drop a call to clone, use iter directly

runcom:
iter returns a reference, ok to call it in the for just like this


nullr0ute (approved):
Looks good

PR #306, opened 2022-08-02
fix(passwordless_login): login with ssh key only

say-paul:
for security reason login with password is disabled for user created
during device onboarding
user created will able to login only using ssh key passed via
service info config.

Signed-off: Sayan Paul <paul.sayan@gmail.com>


7flying:
@say-paul apparently this breaks the e2e test, can you take a look?

PR #308, opened 2022-08-05
Update example rendezvous-server.yml

paulwhalen:
Update rendezvous-server.yml to include 'max_wait_seconds:' and updated storage path used in the AIO. This prevents a panic when trying to run the server without 'max_wait_seconds'.

This also removes the 'trusted_device_keys_path: /path/to/keys/device_ca_cert.pem', as it doesn't seem to be needed (not included in the config when running the fdo-aio). Please confirm before merging. 

Signed-off-by: Paul Whalen <pwhalen@redhat.com>

### DIFF
--- a/examples/config/rendezvous-server.yml
+++ b/examples/config/rendezvous-server.yml
@@ -1,9 +1,10 @@
 ---
 storage_driver:
   Directory:
-    path: /path/to/rendezvous_registered/
-session_store_driver: Directory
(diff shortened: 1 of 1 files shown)


nullr0ute:
can you add "chore: "to the first line and the comment you have for the PR in the actual commit message

nullr0ute (approved):
LGTM

PR #325, opened 2022-10-04
store: add metadata file support

7flying:
Adds metadata file support for systems that
do not support xattrs.

Integration tests for both the xattr storage
mode and the new metadata file storage mode.

Marked as draft because:
- [ ] I haven't added a test yet for every function of `DirectoryStore` and `DirectoryStoreFilterType`. Also the tests are both for the  xattr storage mode and the new metadata file storage mode.
- [ ] The tests need to feed on a standard config file, which I haven't parametrised yet
- [ ] I need to take another look at the code.
- [ ] CI might complain.

### DIFF
--- a/store/src/directory.rs
+++ b/store/src/directory.rs
@@ -1,8 +1,12 @@
+use regex;
+use std::collections::HashMap;
 use std::collections::HashSet;
 use std::convert::TryInto;
 use std::fs::{self, File};
 use std::marker::PhantomData;
 use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::str::{self, FromStr};
 use std::time::{Duration, SystemTime};
 
 use async_trait::async_trait;
@@ -10,8 +14,10 @@ use xattr::FileExt;
 
 use fdo_data_formats::Serializable;
 
-use crate::{FilterType, MetadataLocalKey, MetadataValue, ValueIter};
+use crate::{format_xattr, FdoMetadata, FilterType, MetadataLocalKey, MetadataValue, ValueIter};
+use crate::{set_metadata_extension_to_path, FDO_METADATA_EX};
 
+use super::DirectoryStorageMode;
 use super::Store;
 use super::StoreError;
 
@@ -43,11 +49,55 @@ where
         ))
     })?;
 
+    let xattr_support = check_xattr_support(&canonicalized_directory)?;
+
     Ok(Box::new(DirectoryStore {
         phantom_k: PhantomData,
         phantom_v: PhantomData,
 
         directory: canonicalized_directory,
+        xattr_enabled: xattr_support,
+    }))
+}
+
+pub(super) fn initialize_explicit_mode<OT, K, V, MKT>(
+    path: &Path,
+    mode: DirectoryStorageMode,
+) -> Result<Box<dyn Store<OT, K, V, MKT>>, StoreError>
+where
+    OT: crate::StoreOpenMode + 'static,
+    K: std::str::FromStr + std::string::ToString + Send + Sync + 'static,
+    V: Serializable + Send + Sync + Clone + 'static,
+    MKT: crate::MetadataLocalKey + 'static,
+{
+    if !path.is_absolute() {
+        return Err(StoreError::Configuration(
+            "Storage directory is not absolute".to_string(),
+        ));
+    }
+    fs::create_dir_all(path).map_err(|e| {
+        StoreError::Configuration(format!(
+            "Storage directory '{:?}' could not be created: {}",
+            path, e
+        ))
+    })?;
+
+    let canonicalized_directory = path.canonicalize().map_err(|e| {
+        StoreError::Configuration(format!(
+            "Storage directory '{:?}' could not be canonicalized: {}",
+            path, e
+        ))
+    })?;
+    let xattr_enabled = match mode {
+        DirectoryStorageMode::MetadataFile => false,
+        DirectoryStorageMode::Xattr => true,
+    };
+    Ok(Box::new(DirectoryStore {
+        phantom_k: PhantomData,
+        phantom_v: PhantomData,
+
+        directory: canonicalized_directory,
+        xattr_enabled,
     }))
 }
 
@@ -57,6 +107,7 @@ struct DirectoryStore<K, V> {
     phantom_v: PhantomData<V>,
 
     directory: PathBuf,
+    xattr_enabled: bool,
 }
 
 impl<K, V> DirectoryStore<K, V>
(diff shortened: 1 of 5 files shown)


store/src/directory.rs
```diff
@@ -68,6 +119,101 @@ where
...
+            "reiser" => Ok(FileSystemXattrSupport::Reiserfs),
+            "squashfs" => Ok(FileSystemXattrSupport::Squashfs),
+            "ubifs" => Ok(FileSystemXattrSupport::Ubifs),
+            "xfs" => Ok(FileSystemXattrSupport::Xfs),
+            "yaffs2" => Ok(FileSystemXattrSupport::Yaffs2), // check
+            "zfs" => Ok(FileSystemXattrSupport::Zfs),
+            _ => Err(StoreError::Configuration(format!("Unsupported filesystem"))),
+        }
+    }
+}
```
runcom, line 166: so, usually, I think a better approach is "testing for features" - so instead of hardcoding this info here, you can just try and set xattrs, if the doesn't work we can safely assume it's not supported

store/src/directory.rs
```diff
@@ -138,41 +281,126 @@ where
...
             let mut neqs: HashSet<PathBuf> = HashSet::new();
             for neq in &self.neqs {
                 let (key, expected) = neq;
-                match xattr::get(path.clone(), format_xattr(key)) {
-                    Ok(Some(v)) => {
-                        let matching = expected.iter().zip(&v).filter(|&(a, b)| a == b).count();
-                        if expected.len() != matching {
+
+                if self.xattr_enabled {
+                    match xattr::get(path.clone(), format_xattr(key)) {
```
runcom, line 301: what if we further abstract xattrs enabled or not behind a common trait? we can then have a simple interface for querying them and the underlying implementation will make decisions based on whether xattrs are supported or not

store/src/lib.rs
```diff
@@ -236,4 +283,21 @@ impl StoreConfig {
...
+        // K and V are supersets of the possible requirements for the different implementations
+        K: Eq + std::hash::Hash + Send + Sync + std::string::ToString + std::str::FromStr + 'static,
+        V: Send + Sync + Clone + Serializable + 'static,
+        MKT: crate::MetadataLocalKey + 'static,
+    {
+        match self {
+            #[cfg(feature = "directory")]
+            StoreConfig::Directory { path } => directory::initialize_explicit_mode(path, mode),
+        }
+    }
```
runcom, line 302: this kind of seem lot of duplicate code - I feel everything around xattrs support should really just be hidden within the directory code w/o touching anything else

store/src/directory.rs
```diff
@@ -256,30 +536,94 @@ where
...
+            let file = match File::create(&path) {
+                Ok(f) => f,
+                Err(e) => {
+                    return Err(StoreError::Unspecified(format!(
+                        "Error opening file (w): {}",
+                        e
+                    )))
+                }
+            };
+            match serde_cbor::to_writer(&file, &metadata) {
```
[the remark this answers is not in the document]
nullr0ute: The spec uses cbor explicitly

store/src/lib.rs
```diff
@@ -236,4 +283,21 @@ impl StoreConfig {
             StoreConfig::Directory { path } => directory::initialize(path),
         }
     }
+
+    pub fn initialize_explicit_mode<OT, K, V, MKT>(
```
runcom, line 287: I see this is for testing actually which makes sense but I'd love if we could still find an elegant solution to this (including abstracting the if/else conditionals on whether xattrs are supported or not) (I'll leave it to you Irene to explore this)

store/src/directory.rs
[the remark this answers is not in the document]
runcom: this is not anything spec related afaict, this is how we store metadata and storing it in cbor really makes things difficult to debug if not strictly necessary

store/src/directory.rs
```diff
@@ -392,21 +812,80 @@ where
...
-            if SystemTime::now() < ttl {
-                continue;
-            }
-            log::trace!("File at {} has expired, attempting removal", path.display());
-            if let Err(e) = fs::remove_file(&path) {
-                log::info!("Error deleting expired file {}: {}", path.display(), e);
+                log::trace!("File at {} has expired, attempting removal", path.display());
+                if let Err(e) = fs::remove_file(&path) {
+                    log::info!("Error deleting expired file {}: {}", path.display(), e);
+                }
```
runcom, line 888: this remove_file is basically duplicate with the other in !xattr_enabled so just move it out of the if/else at the end

store/src/directory.rs
```diff
@@ -351,15 +745,41 @@ where
...
+                    return Err(StoreError::Unspecified(format!(
+                        "Error removing associated metadata file '{}': {:?}",
+                        path.display(),
+                        e,
+                    )))
+                }
+            }
+        } else {
+            Ok(())
+        }
```
runcom, line 778: you don't need the if/else here - removing this file and logging if there was an error (not a not found) would imply the use or xattr or not (also remove the return Err as it's the last statement anyway)
```rust
        match fs::remove_file(&set_metadata_extension_to_path(&path)?) {
            Ok(_) => Ok(()),
            Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(()),
            Err(e) => Err(StoreError::Unspecified(format!(
                "Error removing associated metadata file '{}': {:?}",
                path.display(),
                e,
            ))),
        }
```

store/src/lib.rs
```diff
@@ -216,12 +218,57 @@ pub trait Store<OT: StoreOpenMode, K, V, MKT: MetadataLocalKey>: Send + Sync {
...
+}
+
+/// Format of FDOmetadata
+#[derive(Debug, Serialize, Deserialize)]
+pub struct FdoMetadata {
+    pub map: HashMap<String, Vec<u8>>,
+}
+
+/// extension used in our metadata files
+pub const FDO_METADATA_EX: &str = "fdo-md";
```
runcom, line 244: just drop `fdo-` as this file is gonna live within an fdo context anyway (not even meant to be passed around)

store/src/directory.rs
```diff
@@ -392,21 +812,80 @@ where
...
+            // skip non-metadata files if xattr is not supported
+            if !self.xattr_enabled {
+                match path.as_path().extension() {
+                    None => continue,
+                    Some(ex) => {
+                        if ex != FDO_METADATA_EX {
+                            continue;
+                        }
+                    }
+                }
```
runcom, line 824: metadata mgmt shouldn't filter here, we still want to "play" on OVs, the metadata part should be hidden

store/src/directory.rs
```diff
@@ -392,21 +812,80 @@ where
...
+                let path_meta = set_metadata_extension_to_path(&path)?;
```
runcom, line 825: uhm, if the paths here are _only_ metadata files, why doing this? and also, when removing the ov and metadata below, how do you actually remove the OV if we get here only if it's metadata file

store/src/directory.rs
```diff
@@ -392,21 +812,80 @@ where
...
+                    }
+                    None => {
+                        continue;
+                    }
+                };
+                log::trace!("File at {} has expired, attempting removal", path.display());
+                if let Err(e) = fs::remove_file(&path) {
+                    log::warn!("Error deleting expired file {}: {}", path.display(), e);
+                }
+                if let Err(e) = fs::remove_file(&path_meta) {
```
runcom, line 863: path and path_meta are the same here... as we're filtering out non-metadata files

store/src/directory.rs
```diff
@@ -392,21 +812,80 @@ where
                 Ok(v) if v.is_file() => {}
                 Ok(_) => continue,
             }
-            let ttl = match xattr::get(&path, format_xattr(crate::MetadataKey::<MKT>::Ttl.to_key()))
-            {
-                Err(e) => {
-                    log::trace!("Error looking up TTL xattr for {}: {:?}", path.display(), e);
+            // skip non-metadata files if xattr is not supported
+            if !self.xattr_enabled {
```
runcom, line 816: ultimately, we want to remove this if statement here and have something more generic as:
```rust
+fn get_ov_ttl(path: &PathBuf, md_key: &str, xattr_enabled: bool) -> Option<SystemTime> {
+    // we should only ever play with OVs, metadata is an impl detail
+    match path.as_path().extension() {
+        None => (),
+        Some(ex) => {
+            if ex == FDO_METADATA_EX {
+                return None;
+            }
+        }
+    }
+    if xattr_enabled {
+        match xattr::get(&path, format_xattr(md_key)) {
+            Err(e) => {
+                log::trace!("Error looking up TTL xattr for {}: {:?}", path.display(), e);
+                return None;
+            }
+            Ok(None) => return None,
+            Ok(Some(val)) => match ttl_from_disk(&val) {
+                Ok(ttl) => return Some(ttl),
+                Err(e) => {
+                    log::trace!("error converting ttl from disk {:?}", e);
+                    return None;
+                }
+            },
+        };
+    }
+    let path_meta = set_metadata_extension_to_path(&path)?;
+    ......
+    ......
+    None
+}
```
The same goes for the other functions doing if/else

store/src/directory.rs
7flying: I wanted to avoid that in order to touch the minimum number of files, but I guess that switching on a flag is not that clean. Will do that.

store/src/directory.rs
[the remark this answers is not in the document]
7flying: I simply used cbor because it is what we were using due to the standard everywhere else, so only due to consistency reasons. If we want the metadata files to be readable we could use json.

store/src/directory.rs
7flying: In order to perform the maintenance we need to get the ttl and that's on the metadata file, that's why need need those

nullr0ute:
Closing, going to move to DB storage, can always be reopened.

PR #326, opened 2022-10-04
chore: bump reqwest to 0.11.9

nullr0ute:
There's a build issue with older versions of reqwest so bump to a release we know works.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>


nullr0ute:
We used 0.11.9 in the vendored build in RHEL so we know that works, the 0.11.3 does not, so hence the bump so we explicitly use a working release

runcom (approved):

PR #332, opened 2022-10-11
chore: replace chrono crate with time crate

### LINKED ISSUE #329: migrate fdo-admin-tool from chrono to time
The chrono tool has had some issues around maintenance and CVEs and the rest of the bits of the project use time (0.3.15+) so we should look to see if we can migrate to time to remove the dependency on chrono moving forward.

7flying:
Fixes  #329


nullr0ute (approved):
LGTM

PR #335, opened 2022-10-13
chore: push aio image to quay

runcom:

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -68,3 +68,12 @@ jobs:
         tags: ${{ steps.build.outputs.tags }}
         username: ${{ secrets.QUAY_USERNAME }}
         password: ${{ secrets.QUAY_ROBOT_TOKEN }}
+
+    - name: Push aio to quay.io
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -68,3 +68,12 @@ jobs:
         tags: ${{ steps.build.outputs.tags }}
         username: ${{ secrets.QUAY_USERNAME }}
         password: ${{ secrets.QUAY_ROBOT_TOKEN }}
+
+    - name: Push aio to quay.io
+      uses: redhat-actions/push-to-registry@v2
+      with:
+        registry: "quay.io/${{ env.quay_org }}"
+        image: aio
```
nullr0ute, line 76: do we want this to be called something like fdo-aio?

runcom: last time with patrick we said fdo was a repetition - perhaps `aio-server`?

nullr0ute: if the org is fdo I think just aio is fine

nullr0ute (approved):
LGTM

PR #341, opened 2022-10-24
docs: update HOWTO with new Device Credential location

7flying:
New device credential look up location was added in #204.


nullr0ute (approved):
LGTM

PR #342, opened 2022-10-25
chore: port owner-tool to clap v3.1

### LINKED ISSUE #327: migrate owner-tool to user clap v3.1
The owner-tool uses clap v2 but fdo-admin-tool uses clap v3.1 so we need to migrate the owner-tool to use the same version so we don't have different bits of the code using different versions/APIs

miles170:
Migrate `owner-tool` to clap v3.1, with the derive format.

Closes #327.


7flying (approved):
LGTM, thanks for your contribution!

PR #344, opened 2022-10-31
rendezvous-info.yml: fix syntax

muvarov:
Wrong syntax in the current example file. Fxing it.

Signed-off-by: Maxim Uvarov <maxim.uvarov@linaro.org>


7flying (dismissed):
Good catch!

7flying:
You'll need to update your commit message, you can change it to `fix: syntax in sample rendezvous-info.yml` for instance.

7flying (approved):
LGTM!

PR #348, opened 2022-11-09
build: client-linuxapp: remove redurant openssl dependency

muvarov:
No need to directly specify openssl dependency here.

Signed-off-by: Maxim Uvarov <maxim.uvarov@linaro.com>


7flying (approved):
LGTM, thanks!

PR #351, opened 2022-11-14
docs: fix refs to device-credentials

muvarov:
Current implementation takes /etc/device-credentials as env document. Docs in some places use "_" instead of "-" and in some places "credential" instead of "credentials".

Signed-off-by: Maxim Uvarov <maxim.uvarov@linaro.org>


7flying:
I'm sorry but I see no issue with the current docs since they align with what the tool says.
If you believe that the command-line tool `fdo-owner-tool initialize-device` should use `device-credentials-out` instead of `device-credential-out` and thereby the examples should be in singular please open an issue to discuss it.

PR #357, opened 2022-12-16
chore: bump clap to 4.0+

7flying:
Bumps clap to the latest version


Cargo.lock
```diff
@@ -250,7 +250,7 @@ version = "0.24.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a6358dedf60f4d9b8db43ad187391afe959746101346fe51bb978126bec61dfb"
 dependencies = [
- "clap",
+ "clap 3.2.23",
```
nullr0ute, line 253: I would have thought this would have gone to 4.0.x not 3.2.x wit this bump, I suspect it just needs a regen again post the latest changes.

7flying: That's a dependency of `cbindgen`, we are still using 4.0.x.
At the bottom of the Cargo.lock you can see that we have both 3.2.23 and 4.0.29 as our dependencies because of this.

PR #368, opened 2022-12-21
chore: bump nix to 0.26

7flying:
Bumps nix to ~~0.24~~ 0.26, replace deprecated methods/structs.


7flying:
Note that in the Cargo.lock file we list two versions of `nix` as dependencies since there are some of our dependencies that haven't been updated to the latest version yet (and we cannot do anything about that).

nullr0ute (approved):
LGTM

Issue #333, opened 2022-10-12
Support for device identification other than serial number


jritter:
With the current implementation, matching a ownership voucher to a device only seems to be possible through the serial number. It would be great to be able to use other attributes, for instance the MAC address of the first interface.

In my experience, some vendors do a really bad job writing the serial number of their hardware to the SMBIOS, the possibility to find a uniquely identifiable MAC address for a device is much more likely.

nullr0ute:
The problem with MACs is they're often changeable/movable.

jritter:
Agree, but I'd argue that serial numbers are not perfect either. I'm currently working on a project with embedded computers where the manufacturer doesn't write the serial number into the SMBIOS. The MAC address of the embedded NIC would be a reasonable alternative in this case IMHO.

And if this was configurable the user can still choose which one is the better option for the use case at hand.

nullr0ute:
> Agree, but I'd argue that serial numbers are not perfect either. I'm currently working on a project with embedded computers where the manufacturer doesn't write the serial number into the SMBIOS. The MAC address of the embedded NIC would be a reasonable alternative in this case IMHO.

Is the device using U-Boot in this case? The SystemReady-IR standard is working to ensure that is provided as part of certified devices.  

> And if this was configurable the user can still choose which one is the better option for the use case at hand.

How would the user choose that?

The "proper" fix to this is to use device credentials from a TPM2 or other similar HW security device.

Also can you link to where it's using serial # from  SMBIOS in the code?

jritter:
> Is the device using U-Boot in this case? The SystemReady-IR standard is working to ensure that is provided as part of certified devices.

No, it is an Advantech x86_64 single board computer that unfortunately comes without a TPM. We are planning to deploy RHEL for Edge 9 on it. For booting we use UEFI and Grub.

> How would the user choose that?

I'd say in the configuration of the manufacturing server:
https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/examples/config/manufacturing-server.yml#L26

> The "proper" fix to this is to use device credentials from a TPM2 or other similar HW security device.

In an ideal world I agree. Unfortunately we don't have that in our scenario. 

> Also can you link to where it's using serial # from SMBIOS in the code?
https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/manufacturing-client/src/main.rs#L300-L302

[...]

PR #396, opened 2023-01-24
 Extend support for device identification other than serial number

sarmahaj:
This PR contains changes that:
- extends support for device-identification for edge devices other than serial number (which is supposed to be provided by manufacturer ).
- But if serial number from manufacturer (/sys/devices/virtual/dmi/id/product_serial or /sys/devices/virtual/dmi/ id/chassis_serial) is not available then it will be populated as mac addresses of available network interfaces like :
Device Info: IFACE_eth0=02:42:ac:11:00:03;IFACE_eth1=...and so 
-if serial number is present then
Device Info: SNXXXXX;IFACE_eth0=02:42:ac:11:00:03;IFACE_eth1=...and so
- Manual testing done and Device Info in OV gets updated as above.

Signed-off-by: Sarita Mahajan <sarmahaj@redhat.com>


manufacturing-client/src/main.rs
```diff
@@ -291,15 +293,57 @@ async fn main() -> Result<()> {
 }
 
 async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType> {
-    if let Some(mfg_info) = env::var_os("MANUFACTURING_INFO") {
-        return Ok(CborSimpleType::Text(mfg_info.into_string().unwrap()));
-    }
     match mfg_string_type {
         MfgStringType::SerialNumber => {
```
7flying, line 297: I think that we could define a `MfgStringType::MAC` type so that people can directly specify if they want the MAC address as the identifier in the `manufacturing_server.yml` config file. The MAC would still be the fallback when no serial number is found, but giving the users this explicit option would make things more clear.

7flying:
Fixes #333

manufacturing-client/src/main.rs
```diff
@@ -291,15 +293,57 @@ async fn main() -> Result<()> {
 }
 
 async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType> {
-    if let Some(mfg_info) = env::var_os("MANUFACTURING_INFO") {
```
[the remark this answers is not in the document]
sarmahaj: Well this is a bit knotty !  so this condition checks if mfs_info is present and equals to MANUFACTURING_INFO , then it sets the mfg_info as return value .This case is always true since we set that to   .env("MANUFACTURING_INFO", &mfg_args.device_info) in here -> https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/admin-tool/src/aio/device.rs#L105. Therefore in OV Device_Info always reflects as the device name which passed in the command when device is created like target/debug/fdo-admin-tool aio --directory aio-dir device --device-credential-location testdevice1.dc manufacture --device-info testdevice1 . So if we keep this condition , testdevice1 appears as msf_info always. And never goes to check Serial Number (below match condition)
May be this needs some refactoring how the get_mfg_info() is written or a separate function for getting mfg_info and serialNumber.

manufacturing-client/src/main.rs
```diff
@@ -291,15 +293,57 @@ async fn main() -> Result<()> {
...
+                .arg("/sys/class/net/")
+                .stdout(std::process::Stdio::piped())
+                .stderr(std::process::Stdio::piped())
+                .output()
+                .expect("Failed to execute command");
+
+            let val = String::from_utf8_lossy(&output_ifaces.stdout).to_string();
+            let mut serial_mac: String = "".to_string();
+            for interface in val.split('\n') {
+                if interface.starts_with('e') || interface.starts_with('w') {
```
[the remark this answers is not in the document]
sarmahaj: I like that option, it will be better way to go through all network interfaces without concerning the name (even if it changes by some means)and get proper mac address.

manufacturing-client/src/main.rs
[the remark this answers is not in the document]
7flying: This only happens with the `aio` tool which is not used in production

manufacturing-client/src/main.rs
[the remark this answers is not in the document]
runcom: yeah, mac(s) can be changed and I agree we should just discard empty mac and loopback

manufacturing-client/src/main.rs
runcom: I think we shouldn't fallback, if the user asks for SerialNumber they get the old behavior - if they ask for the new one they get the new info which may or may not include the SN. I think we should call it something like `MfgStringType::Extended` or something like that, the point is this new format allows us to add fields to the mfgstring in a backwards compatible way too

manufacturing-client/src/main.rs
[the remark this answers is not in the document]
7flying: The developer has never had the OV with a "real" serial number, the env variable is there because we don't need anything real for testing how the servers work, which is the aim of `aio`. And also, we don't need to test everything with `aio`, it is mainly to test thing that require interactions with different servers.

I feel like you can handle this in two ways: 1) do not delete this environment variable, because it does not affect production and `aio` works fine the way that it is (its goal is to spin up the servers for testing quickly, so having a real device info is not relevant) and make a test outside e2e for your new functionality;
or 2) update *all* the e2e tests that require the handling of this environment variable, make your new tests (outside e2e) handling this new functionality, update the `aio` tool so that it can generate the Device Info in three different ways (1: serial number, 2: new behaviour with mac + serial number, 3: a way to overwrite 1 & 2 and simply put in the OV something that we want —which is what the env variable did—), and finally update the `aio` tool tests so that this new functionality of `aio` is tested.

manufacturing-client/src/main.rs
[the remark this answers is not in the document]
sarmahaj: added check for valid and non-zero MAC address in latest commit . so that it will drop empty mac

manufacturing-client/src/main.rs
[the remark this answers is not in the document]
sarmahaj: with the pov that "we don't need to test everything with aio, it is mainly to test thing that require interactions with different servers." I believe then it will be better to add a test outside e2e only to test new format of the serial number. (so that that avoids changes in all previous test cases and this new functionality gets tested anyways in separate test case). I am working on adding test case to this.

data-formats/src/constants/mod.rs
```diff
@@ -425,15 +425,15 @@ pub enum ErrorCode {
...
+    SerialNumberExt = 0,
 }
 
 impl FromStr for MfgStringType {
     type Err = Error;
 
     fn from_str(s: &str) -> Result<Self> {
         Ok(match &s.to_lowercase()[..] {
-            "serialnumber" | "serial_number" => MfgStringType::SerialNumber,
+            "serialnumber" | "serial_number" => MfgStringType::SerialNumberExt,
```
runcom, line 436: I think _changing_ this as opposed as _adding_ a new config option would be best - otherwise we risk people finding out we introduced new unexpected behavior for some

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
                 .or_else(|_| fs::read_to_string("/sys/devices/virtual/dmi/id/chassis_serial"))
                 .context("Error determining system serial number")?;
-            Ok(CborSimpleType::Text(serial))
+            let mut arg: String;
+            let output_ifaces = Command::new("ls")
+                .arg("/sys/class/net/")
+                .stdout(std::process::Stdio::piped())
+                .stderr(std::process::Stdio::piped())
+                .output()
+                .expect("Failed to execute command");
```
cgwalters, line 311: `?`

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+
```
cgwalters, line 312: Probably want to
```rust
   let st = output_ifaces.status;
   if !st.success() {
      return anyhow::bail!("Failed to execute ls: {st:?}");
   }
```

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+            let val = String::from_utf8_lossy(&output_ifaces.stdout).to_string();
+            let mut serial_mac: String = "".to_string();
+            for interface in val.split('\n') {
```
cgwalters, line 315: Can use `.lines()` here.

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                arg = "/sys/class/net/".to_owned() + interface + "/address";
```
cgwalters, line 316:
```suggestion
                let arg = format!("/sys/class/net/{interface}/address");
```

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                let output = Command::new("/bin/cat")
+                    .args([":", &arg])
+                    .stdout(std::process::Stdio::piped())
+                    .stderr(std::process::Stdio::piped())
+                    .output()
+                    .expect("Failed to execute grep command");
+                let mac = String::from_utf8_lossy(&output.stdout)
+                    .trim_end()
+                    .to_string();
+                let mac_regex = Regex::new(r"^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$").unwrap();
```
cgwalters, line 326: Often want to cache regexps; check out https://github.com/coreos/coreos-installer/pull/1101

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                if mac_regex.is_match(&mac) {
+                    let mut non_zero_mac = false;
+                    for byte in mac.split(':') {
+                        let byte = u8::from_str_radix(byte, 16).unwrap();
```
cgwalters, line 330: I'd probably use `?` here too.

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+            let mut arg: String;
+            let output_ifaces = Command::new("ls")
+                .arg("/sys/class/net/")
+                .stdout(std::process::Stdio::piped())
+                .stderr(std::process::Stdio::piped())
+                .output()
+                .expect("Failed to execute command");
+
+            let val = String::from_utf8_lossy(&output_ifaces.stdout).to_string();
+            let mut serial_mac: String = "".to_string();
```
cgwalters, line 314:
```rust
            let mut serial_mac = String::new();
```

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                    let mut non_zero_mac = false;
+                    for byte in mac.split(':') {
+                        let byte = u8::from_str_radix(byte, 16).unwrap();
+                        if byte != 0 {
+                            non_zero_mac = true;
+                            break;
+                        }
+                    }
+                    if non_zero_mac {
+                        serial_mac = serial_mac
```
cgwalters, line 337: Maybe a bit more ergonomic:

```
                        write!(serial_mac, "IFACE_{interface}={mac};").unwrap();
```

I'd unwrap here because this can't actually fail (except on OOM) but Rust panic on that anyways.

manufacturing-client/src/main.rs
cgwalters: Another thing that'd be good to do here is to trim off the newline:
```
                 let interface = interface.trim();
```

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                            .add(";");
+                        log::info!(
+                            "MAC address {:#?} of interface {:#?} \n",
+                            serial_mac,
+                            interface
+                        );
+                    }
+                }
+            }
+            if serial_dmi == "None\n" || serial_dmi.is_empty() {
```
cgwalters, line 351: Then this can use `if serial_dmi == "None"` with the newline trimmed.

manufacturing-client/src/main.rs
```diff
@@ -295,11 +298,70 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                .stdout(std::process::Stdio::piped())
+                .stderr(std::process::Stdio::piped())
+                .output()
+                .expect("Failed to execute command");
+
+            let val = String::from_utf8_lossy(&output_ifaces.stdout).to_string();
+            let mut serial_mac: String = "".to_string();
+            for interface in val.split('\n') {
+                arg = "/sys/class/net/".to_owned() + interface + "/address";
+                let output = Command::new("/bin/cat")
```
cgwalters, line 317: I think this can just use `fs::read_to_string()` too.

manufacturing-client/src/main.rs
cgwalters: (Also now that I think about it, instead of forking `ls` it's likely better to use https://doc.rust-lang.org/std/fs/fn.read_dir.html )

data-formats/src/constants/mod.rs
sarmahaj: taken care in latest commit , now user can specify SerialNumber or SerialNumberExt value for mfg_string_type in manufacturing_server.yml , which will decide existing(old) or the new format for serial number

manufacturing-client/src/main.rs
sarmahaj: Thanks for the suggestions , taken care in the latest commit !

manufacturing-client/src/main.rs
sarmahaj: great catch !

manufacturing-client/src/main.rs
```diff
@@ -296,10 +301,79 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+            );
+            let serial_dmi = fs::read_to_string("/sys/devices/virtual/dmi/id/product_serial")
                 .or_else(|_| fs::read_to_string("/sys/devices/virtual/dmi/id/chassis_serial"))
                 .context("Error determining system serial number")?;
-            Ok(CborSimpleType::Text(serial))
+            Ok(CborSimpleType::Text(serial_dmi))
+        }
+        MfgStringType::SerialNumberExt => {
+            log::debug!(
+                "msf_string_type value:{:#?} refers SerialNumberExt format ",
```
miabbott, line 315: nit: `s/msf_string_type/msg_string_type/`

miabbott: Or is this really supposed to be `mfg_string_type`?  I'm getting confused by a similar statement on lines 305-306

https://github.com/sarmahaj/fido-device-onboard-rs/blob/343d959b97ec062e306f9d73eb0a527c55e17ec2/manufacturing-client/src/main.rs#L305-L306

I think they should be the same, whichever is correct

manufacturing-client/src/main.rs
```diff
@@ -1,5 +1,10 @@
+use lazy_static::lazy_static;
+use regex::Regex;
+use std::fmt::Write;
+use std::process::Command;
 use std::{convert::TryFrom, fs};
 use std::{convert::TryInto, env, str::FromStr};
+//use std::fs;
```
miabbott, line 7: nit: remove this line?

manufacturing-client/src/main.rs
miabbott: Looks like it is still forking and not using `fs::read_dir`?

manufacturing-client/src/main.rs
```diff
@@ -296,10 +301,79 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                            "MAC address {:#?} of interface {:#?} \n",
+                            serial_mac,
+                            interface
+                        );
+                    }
+                }
+            }
+            if serial_dmi == "None" || serial_dmi.is_empty() {
+                log::info!(
+                    " Serial number of the device is just mac addr: {:#?} ",
```
miabbott, line 365: nit: remove the leading space from the string?

manufacturing-client/src/main.rs
```diff
@@ -296,10 +301,79 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                    serial_mac
+                );
+                Ok(CborSimpleType::Text(serial_mac))
+            } else {
+                serial_mac = serial_dmi.trim_end().to_string() + ";" + &serial_mac;
+                log::info!(
+                    "Serial number of the device is dmi_serial + mac address {:#?} ",
```
miabbott, line 372: This log message is not clear unless you are reading the code, so end users may be confused here.

Assuming I am reading this correctly, it looks like you are reporting the serial number of the device which is a concatenation of the serial number from the DMI and the MAC address.  So I would expand the log message to something like:

`Serial number of the device the concatenation of the serial number from the DMI and the MAC address {:#?}`

manufacturing-client/src/main.rs
```diff
@@ -296,10 +301,79 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+
+            let val = String::from_utf8_lossy(&output_ifaces.stdout).to_string();
+            let mut serial_mac = String::new();
+            for interface in val.lines() {
+                let path_mac_addr = format!("/sys/class/net/{interface}/address");
+                let mac = fs::read_to_string(path_mac_addr)
+                    .map(|s| s.trim_end().to_owned())
+                    .context("Error reading mac address")?;
+                log::info!(
+                    "mac read from file /sys/class/net/{interface}/address {:#?} ",
```
miabbott, line 337: nit: a little more verbosity + capitlization:

`MAC address read from file /sys/class/net/{interface}/address {:#?}`

manufacturing-client/src/main.rs
miabbott: nit: `s/mac addr/MAC address/`

manufacturing-client/Cargo.toml
```diff
@@ -16,6 +16,8 @@ sys-info = "0.9"
 passwd = "0.0.1"
 rand = "0.8.4"
 tss-esapi = "7.2"
+regex = "1.3.7"
+lazy_static = "^1.4"
```
7flying, line 20: Why the `^`? `1.4` should be OK

manufacturing-client/src/main.rs
7flying: `mfg` is the shortened version of `manufacturer` <https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#protocolEnt>, it should be `mfg`

manufacturing-client/src/main.rs
7flying: I do agree, please use `fs:read_dir`

manufacturing-client/src/main.rs
```diff
@@ -296,10 +301,79 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                    mac
```
7flying, line 338: if you're using inlined format args with `log::info!` continue to do so with the rest of the parameters: `log::info!(" ...{interface} ... {mac:#?");`

manufacturing-client/src/main.rs
```diff
@@ -296,10 +301,79 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+                        let byte = u8::from_str_radix(byte, 16)?;
+                        if byte != 0 {
+                            non_zero_mac = true;
+                            break;
+                        }
+                    }
+                    if non_zero_mac {
+                        write!(serial_mac, "IFACE_{interface}={mac};").unwrap();
+                        log::debug!(
+                            "MAC address {:#?} of interface {:#?} \n",
```
7flying, line 356: you can also use inline format args here if you fancy so

7flying: there is no need for the \n

manufacturing-client/src/main.rs
```diff
@@ -296,10 +301,79 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
...
+        }
+        MfgStringType::SerialNumberExt => {
+            log::debug!(
+                "msf_string_type value:{:#?} refers SerialNumberExt format ",
+                mfg_string_type
+            );
+            let serial_dmi = fs::read_to_string("/sys/devices/virtual/dmi/id/product_serial")
+                .or_else(|_| fs::read_to_string("/sys/devices/virtual/dmi/id/chassis_serial"))
+                .map(|s| s.trim_end().to_owned())
+                .context("Error determining system serial number")?;
```
7flying, line 321: It might make sense to trim `serial_dmi` over here

util/src/servers/configuration/manufacturing_server.rs
```diff
@@ -59,13 +59,15 @@ pub struct DiunSettings {
 
 #[derive(Debug, Serialize, Deserialize, Clone, Copy)]
 pub enum MfgStringTypeString {
-    SerialNumber,
+    SerialNumber = 0,
+    SerialNumberExt = 1,
```
7flying, line 63: There is no need to assign a constant to the enum, if we need to print them then it makes more sense to implement the [std::fmt::Display](https://doc.rust-lang.org/std/fmt/trait.Display.html#) trait

integration-tests/tests/serial_number_test.rs
```diff
@@ -0,0 +1,99 @@
+mod common;
+use anyhow::{bail, Context, Result};
+use common::{Binary, LogSide, TestContext};
+use lazy_static::lazy_static;
+use regex::Regex;
+use std::{io::Read, time::Duration};
+
```
7flying, line 7: ^ as the warning says

integration-tests/tests/serial_number_test.rs
```diff
@@ -0,0 +1,99 @@
...
+const L: LogSide = LogSide::Test;
+
+#[tokio::test]
+async fn test_device_info_serialnumber_ext() -> Result<()> {
+    let mut ctx = TestContext::new().context("Error building test context")?;
+
+    //Importtant to add cfg serial_number_format as SerialNumberExt to test this feature
```
7flying, line 14: s/Importtant/ Important/

integration-tests/tests/serial_number_test.rs
```diff
@@ -0,0 +1,99 @@
...
+                        Regex::new(r"^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$").unwrap();
+                }
+
+                for item in line_str.split(';') {
+                    if let Some(iface) = item.strip_prefix("IFACE_") {
+                        let parts: Vec<&str> = iface.split('=').collect();
+                        let mac_address = parts[1];
+                        if MAC_REGEX.is_match(mac_address) {
+                            L.l(format!(
+                                "Found valid MAC address: test_device_info_serial_number passes{}",
```
7flying, line 83: nit: space before the {},

manufacturing-client/src/main.rs
```diff
@@ -296,10 +301,79 @@ async fn get_mfg_info(mfg_string_type: MfgStringType) -> Result<CborSimpleType>
     }
     match mfg_string_type {
         MfgStringType::SerialNumber => {
-            let serial = fs::read_to_string("/sys/devices/virtual/dmi/id/product_serial")
+            log::debug!(
+                "msg_string_type value:{:#?} refers SerialNumber format ",
```
7flying, line 305: mfg

integration-tests/tests/serial_number_test.rs
```diff
@@ -0,0 +1,99 @@
...
+        .run_owner_tool(
+            client_result.client_path(),
+            &["dump-device-credential", dc_path.to_str().unwrap()],
+        )
+        .context("Error running dump-device-credential")?;
+    owner_output
+        .expect_success()
+        .context("Dump-device-credential failed")?;
+
+    for line in owner_output.raw_stdout().split(|&b| b == b'\n') {
```
7flying, line 69: if the `owner_output` is empty for some reason the test will exit with the `Ok(())`, we should fail in that case

7flying (changes requested):
We do have the two ways of identifying the devices now, plus tests, I've added some comments

integration-tests/tests/serial_number_test.rs
sarmahaj: taken care here

manufacturing-client/src/main.rs
sarmahaj: done ,thanks for suggesting !

manufacturing-client/src/main.rs
sarmahaj: thanks , log msg updated

manufacturing-client/src/main.rs
sarmahaj: added , thanks

util/src/servers/configuration/manufacturing_server.rs
sarmahaj: agreed there is no specific need for assigned values (had added it for testing so removed it) , thanks.

manufacturing-client/src/main.rs
sarmahaj: yes ,it does not need trim later in the code . Its done here

data-formats/src/constants/mod.rs
```diff
@@ -426,6 +426,7 @@ pub enum ErrorCode {
 #[non_exhaustive]
 pub enum MfgStringType {
     SerialNumber = 0,
+    SerialNumberExt = 1,
```
runcom, line 429: We can do a bit of bikeshedding here about naming (which is hard anyway but still). This new format provide a structured device identification so _maybe_ we could name it `Structured` or `fill_in_your_suggestion` - the reason for not using `SerialNumberExt` is because it doesn't really tell what `Ext` is and can be just as weird to then look at it and see mac addresses.

sarmahaj: I have two suggestions , that can give a bit more clarity on what these two types do: 
`pub enum MfgStringType {
                  MfgProvidedSerialNumber,
                  CustomSerialNumberWithMAC }`
or CustomSerialNumberWithMAC -> StructuredSerialNumberWithMAC

data-formats/src/constants/mod.rs
```diff
@@ -425,7 +425,8 @@ pub enum ErrorCode {
 #[repr(i8)]
 #[non_exhaustive]
 pub enum MfgStringType {
-    SerialNumber = 0,
+    SerialNumber,
```
7flying, line 428: This is a different enum from the one at `util/.../manufacturing_server.rs` and it does need the constants.

nullr0ute: Can we centralise them in an include file rather than duplicating?

7flying: No because the ones in `data-formats` are going to be sent over network and the serialization is different from the ones at `util`, which are serialized into files (for user configuration).

data-formats/src/constants/mod.rs
runcom: First one, the current, we can't change w/o losing backwards compatibility. The second one as you suggested Irene is tricky as we might add other things than serial number and mac since it's just a key=value structure. uhm

data-formats/src/constants/mod.rs
sarmahaj: > This is a different enum from the one at `util/.../manufacturing_server.rs` and it does need the constants.

ahh right! since we are reading values of these enum members directly without initializing we need constants! Learned this new that Rust wont assign values by default if not initialised. But I wonder why compiler didnt give any error?

data-formats/src/constants/mod.rs
sarmahaj: right.. how about , SerialNumberExt -> DeviceInfo or DeviceInfoStructure ? As you mentioned this field could be MAC address or any other info of the device which we might support later.
If we need a bit more specific related to device identification then it could be something like : 
DeviceIdInfo .
Also this will relate to the Field is same as we use in device credential & ov to represent this info .

data-formats/src/constants/mod.rs
7flying: > > This is a different enum from the one at `util/.../manufacturing_server.rs` and it does need the constants.
> 
> ahh right! since we are reading values of these enum members directly without initializing we need constants! Learned this new that Rust wont assign values by default if not initialised. But I wonder why compiler didnt give any error?

The constants are needed because of the `#[repr(i8)]` which forces the enum to have the same data layout representation as it would have in C. If we don't set the constant there might be some undefined behaviour going on.

data-formats/src/constants/mod.rs
miabbott: how about `SerialNumberExt` -> `MacDerivedSerialNumber`?

runcom: The issue I have with mac and serial number in the new schema is that it won't really be tied to just those two (if any new requirements comes in the future, say people want the hostname?) - how about `StructuredDeviceInfo`?

sarmahaj: StructuredDeviceInfo sounds good to me, we can go ahead with that if everyone agrees !

sarmahaj: Great ,thanks all!

7flying:
Can you change the commit message to `feat`  from `refactor`?

7flying (approved):
All good, thanks!

PR #401, opened 2023-01-27
ci: add tests for serde_yaml serialization + deserialization

7flying:
Checks the serialization and deserialization of the server configuration files.
This checks that converting Rust types to yaml files and back to Rust types does not break anything.


nullr0ute (approved):
Great to see some more tests.

PR #410, opened 2023-02-03
Reboot option

7flying:
This adds a new serviceinfo api module `reboot` which allows
to specify if the system should be rebooted after onboarding
is completed.

This means that no client errors can be sent back since onboarding
is already considered to be a success. This reboot action is
always handled last, so any other onboarding action may fail
and will be handled according to the protocol. Only if onboarding
is successfully handled the reboot action will take place.

This reboot action is optional and can be configured in the
`serviceinfo-api-server.yml`

### DIFF
--- a/client-linuxapp/src/main.rs
+++ b/client-linuxapp/src/main.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Borrow, env, fs, path::PathBuf, thread, time};
+use std::{borrow::Borrow, env, fs, path::PathBuf, process::Command, thread, time};
 
 use anyhow::{anyhow, bail, Context, Result};
 use rand::Rng;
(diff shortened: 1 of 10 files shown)


runcom:
LGTM - @nullr0ute could you bless this to go in?

client-linuxapp/src/main.rs
```diff
@@ -1218,5 +1226,12 @@ async fn main() -> Result<()> {
         }
     }
     log::info!("Secure Device Onboarding DONE");
+    log::info!("Reboot required? {}", reboot_si_required);
```
nullr0ute, line 1229: what's the _si_ here, and elsewhere, is that meant to be is or is it Spanish for yes, or is it a typo, or something else entirely?

nullr0ute:
> LGTM - @nullr0ute could you bless this to go in?

I have one query, else generally LGTM.

client-linuxapp/src/main.rs
runcom: probably stands for "service info"?

7flying: yep, Service Info

nullr0ute (approved):
LGTM

PR #415, opened 2023-02-07
fix(spec): do not include unpackaged rendezvous-info.yml file

7flying:
Removes rendezvous-info.yml from the install list for the aio command,
the contents of this file are already embedded into the manufacturing-server.yml file,
so aio does not need it.

This fixes the following error when trying to `make rpm` locally:
```
Checking for unpackaged file(s): /usr/lib/rpm/check-files fido-device-onboard-rs/rpmbuild/BUILDROOT/fido-device-onboard-0.4.7-1.20230207gitdaa4a6c.el9.x86_64
error: Installed (but unpackaged) file(s) found:
   /usr/share/doc/fdo/rendezvous-info.yml
```

### DIFF
--- a/fido-device-onboard.spec
+++ b/fido-device-onboard.spec
@@ -70,7 +70,11 @@ install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-admin-to
 install -D -m 0755 -t %{buildroot}%{_bindir} target/release/fdo-owner-tool
 install -D -m 0755 -t %{buildroot}%{_bindir} target/release/fdo-admin-tool
 install -D -m 0644 -t %{buildroot}%{_unitdir} examples/systemd/*
-install -D -m 0644 -t %{buildroot}%{_docdir}/fdo examples/config/*
+# we do not need the rendezvous-info.yml for the AIO command, add everything else
+install -D -m 0644 -t %{buildroot}%{_docdir}/examples examples/config/manufacturing-server.yml
+install -D -m 0644 -t %{buildroot}%{_docdir}/examples examples/config/owner-onboarding-server.yml
+install -D -m 0644 -t %{buildroot}%{_docdir}/examples examples/config/rendezvous-server.yml
+install -D -m 0644 -t %{buildroot}%{_docdir}/examples examples/config/serviceinfo-api-server.yml
 mkdir -p %{buildroot}%{_sysconfdir}/fdo
 # 52fdo
 install -D -m 0755 -t %{buildroot}%{dracutlibdir}/modules.d/52fdo dracut/52fdo/module-setup.sh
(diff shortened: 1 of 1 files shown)


runcom:
should we then just remove the file altogether if it's merged into the manufacturing config and hence unused?

7flying:
> should we then just remove the file altogether if it's merged into the manufacturing config and hence unused?

No because it is referenced in the HOWTO. You need this file in order to generate an OV and a device_credential, so we should provide with an example.

nullr0ute (commented):
Can we drop the fdo and space so it's just %{buildroot}%{_docdir}/examples/ (also no spaces).

7flying:
> Can we drop the fdo and space so it's just %{buildroot}%{_docdir}/examples/ (also no spaces).

Let me know if the latest change makes sense

fido-device-onboard.spec
```diff
@@ -70,7 +70,11 @@ install -D -m 0755 -t %{buildroot}%{_libexecdir}/fdo target/release/fdo-admin-to
 install -D -m 0755 -t %{buildroot}%{_bindir} target/release/fdo-owner-tool
 install -D -m 0755 -t %{buildroot}%{_bindir} target/release/fdo-admin-tool
 install -D -m 0644 -t %{buildroot}%{_unitdir} examples/systemd/*
-install -D -m 0644 -t %{buildroot}%{_docdir}/fdo examples/config/*
+# we do not need the rendezvous-info.yml for the AIO command, add everything else
+install -D -m 0644 -t %{buildroot}%{_docdir}/examples examples/config/manufacturing-server.yml
```
nullr0ute, line 74: you now have two examples and still a space :) so basically "sed -i 's/examples //' fido-device-onboard.spec"

fido-device-onboard.spec
```diff
@@ -103,8 +107,8 @@ Requires: openssl-libs >= 3.0.1-12
 %{_mandir}/man1/fdo-owner-onboarding-server.1*
 %{_libexecdir}/fdo/fdo-owner-onboarding-server
 %{_libexecdir}/fdo/fdo-serviceinfo-api-server
-%{_docdir}/fdo/owner-onboarding-server.yml
-%{_docdir}/fdo/serviceinfo-api-server.yml
+%{_docdir}/examples/owner-onboarding-server.yml
+%{_docdir}/examples/serviceinfo-api-server.yml
```
nullr0ute, line 111: but, yes I think we need this to actually be named %{_docdir}/%{name}/examples/ and same for the ones below.

fido-device-onboard.spec
7flying: The only space that I can see is the space that separates the <dest-dir> from the <source-dir>, what other space is there that I can't see?

nullr0ute: Yep, you're right, I was reading it wrong :-/

nullr0ute (approved):

fido-device-onboard.spec
7flying: No problem!

PR #417, opened 2023-02-15
fix: owner-onboarding service should also start serviceinfoapi service

7flying:
When using systemd units the `fdo-owner-onboarding-server.service` should also start `fdo-serviceinfo-api-server`.


nullr0ute (approved):
LGTM

PR #424, opened 2023-02-17
fix: non-exhaustive patterns on store

7flying:
Fixes a warning on the store


puiterwijk:
I'd specifically not added that, to make sure that adding any new options would require adding the implementation, or it would be a compile-time error. This would make it a run-time error instead.
What warning are you seeing, maybe just allow that lint instead?

7flying:
> I'd specifically not added that, to make sure that adding any new options would require adding the implementation, or it would be a compile-time error. This would make it a run-time error instead. What warning are you seeing, maybe just allow that lint instead?

When we are trying to run an specific test of the `integration-tests` we usually run into this issue, which forces us to use github's CI: 

```
cargo test -p integration-tests --test di_diun-tests

[.....]

   Compiling fdo-store v0.4.10 (/home/idiez/code/repos/fedora-iot/tests/fido-device-onboard-rs/store)
error[E0004]: non-exhaustive patterns: type `&StoreConfig` is non-empty
   --> store/src/lib.rs:234:15
    |
234 |         match self {
    |               ^^^^
    |
note: `StoreConfig` defined here
   --> store/src/lib.rs:220:10
    |
220 | pub enum StoreConfig {
    |          ^^^^^^^^^^^
    = note: the matched value is of type `&StoreConfig`
    = note: references are always considered inhabited
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern as shown
    |
234 ~         match self {
235 +             _ => todo!(),
236 +         }
    |

For more information about this error, try `rustc --explain E0004`.
error: could not compile `fdo-store` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
```

PR #425, opened 2023-02-20
fix: do not start two servers with the same unit

7flying:
We'll go back to the way we used to start `fdo-owner-onboarding-server` and `fdo-serviceinfo-api-server`, with two separate systemd unit files that started each server and that's it.

The .spec file already ships each unit on `fdo-owner-onboarding-server` so on that RPM the user has everything that they need to start each server.


nullr0ute (approved):
LGTM

PR #427, opened 2023-02-21
chore: don't require an explicit minor release of reqwest

nullr0ute:
The reqwest crate is used in a number of places in FDO but it's only specified as minor releases in in the http-wrapper side of things. Let's align major versions as per the rest of the project.


nullr0ute:
It looks like the minor rev bumps here were primarily the auto bot so it should be fine from a rev PoV.

7flying (approved):
LGTM

Issue #321, opened 2022-09-22
Fix user login for initial user


say-paul:
fix for #270 does not solve the issue initial user  created during the onboarding process not able to login as it asks for password which not set via the useradd command.

Probable solution will be to disable password and set login only via cert.

runcom:
@say-paul can you describe what's happening? if I add a user + an ssh key I should be able to just login via the key - can you test this  out now that your PR has merged and report where we fail?

nullr0ute:
If this is the bug I mentioned it's for local login and sudo as you can't use a key for local console based login (think a kiosk or similar) or for sudo so we need to be able to optionally also set a password.

runcom:
> If this is the bug I mentioned it's for local login and sudo as you can't use a key for local console based login (think a kiosk or similar) or for sudo so we need to be able to optionally also set a password.

yes, I'm not questioning the feature itself or the use, it makes complete sense to support this - I wanted to check what is really going on by adding a user and ssh key and then trying to login over ssh

runcom:
again no questioning there :) one issue is surely around the fact that with sayan's #270 I'm still not able to ssh into the system by using the key injected there with that user :) that is something I want to investigate, whether that's an osdefault or not it'll make my suspicion a bug or not

say-paul:
[this](https://github.com/shadow-maint/shadow/issues/222) should be the answer for this issue, though I found out about the default pass today only, but I have tested that it indeed asks for password , which I have never set.

say-paul:
`cat /etc/passwd | grep testuser` shows up as
`testuser:x:1001:1001::/var/home/testuser:/bin/bash` , it creates a entry in /etc/shadow as determined by :x:

runcom:
What I’m asking is “with #270 can you login via ssh at least?”

say-paul:
No, it asks for password instead.

runcom:
I'd go and check on a live system built with IB and using simplified installer to validate this whole flow then

say-paul:
Tested it with the main branch:

1. Ssh works without password.
2. Su - <testuser> asks for password.

runcom:
good, as long as ssh works w/o a password, #270 works as expected and we could continue with this in the future

runcom:
for real-world-like e2e test I would definitely rely on osbuild-composer upstream test where we fully setup the infra to run fdo too - although having something here w/o testing simplified installer is also needed

PR #437, opened 2023-03-02
User creation with password

djach7:
This pr attempts to implement the creation of a user and password when onboarding a device, using a username and password provided within the serviceinfo_api_server.yml config. 

closes https://github.com/fedora-iot/fido-device-onboard-rs/issues/321

This is a draft, current known issues include:
- Password encryption on on user creation
- Confirming sshkey creation still functional
- Confirming edge case functionality

### DIFF
--- a/client-linuxapp/src/serviceinfo.rs
+++ b/client-linuxapp/src/serviceinfo.rs
@@ -77,7 +77,10 @@ fn create_user(user: &str) -> Result<()> {
         log::info!("User {user} created successfully");
         Ok(())
     } else {
-        bail!(format!("User creation failed. Exit Status: {:#?}", status.code()));
+        bail!(format!(
+            "User creation failed. Exit Status: {:#?}",
+            status.code()
+        ));
     }
 }
 
@@ -112,7 +115,6 @@ fn create_user_with_password(user: &str, password: &str) -> Result<()> {
         str_encrypted_pw = sha256_simple(password, &default_params).expect("Hashing failed");
         assert!(sha256_check(password, &str_encrypted_pw).is_ok());
     }
-    
     // Creates new user if user not present
     log::info!("Creating user {user} with password");
     let status = Command::new("useradd")
@@ -128,10 +130,12 @@ fn create_user_with_password(user: &str, password: &str) -> Result<()> {
         log::info!("User {user} created successfully with password");
         Ok(())
     } else {
-        bail!(format!("User creation failed. Exit Status: {:#?}", status.code()));
+        bail!(format!(
+            "User creation failed. Exit Status: {:#?}",
+            status.code()
+        ));
     }
 }
-
 
 fn install_ssh_key(user: &str, key: &str) -> Result<()> {
     let user_info = passwd::Passwd::from_name(user);
@@ -470,21 +474,15 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
         }
         if module == FedoraIotServiceInfoModule::SSHKey.into() {
             if key == "username" {
-                let value = value
-                    .as_str()
-                    .context("Error parsing username value")?;
+                let value = value.as_str().context("Error parsing username value")?;
                 sshkey_user = Some(value.to_string());
                 log::info!("Username is: {value}");
             } else if key == "password" {
-                let value = value
-                    .as_str()
-                    .context("Error parsing password value")?;
+                let value = value.as_str().context("Error parsing password value")?;
                 sshkey_password = Some(value.to_string());
                 log::info!("Password is present");
             } else if key == "sshkeys" {
-                let value = value
-                    .as_str()
-                    .context("Error parsing sshkey value")?;
+                let value = value.as_str().context("Error parsing sshkey value")?;
                 sshkey_keys = Some(value.to_string());
                 log::info!("Keys are present");
             }
@@ -703,10 +701,14 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
         }
         if sshkey_password.is_some() {
             log::info!("SSHkey module was active, creating user with password");
-            create_user_with_password(sshkey_user.as_ref().unwrap(), sshkey_password.as_ref().unwrap()).context(format!(
+            create_user_with_password(
+                sshkey_user.as_ref().unwrap(),
+                sshkey_password.as_ref().unwrap(),
+            )
+            .context(format!(
                 "Error creating new user with password: {}",
-                sshkey_user.as_ref().unwrap() 
-            ))?;  
+                sshkey_user.as_ref().unwrap()
+            ))?;
         }
         if sshkey_keys.is_some() {
             log::info!("SSHkey module was active, installing SSH keys");
@@ -714,10 +716,15 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
                 "Error creating new user: {}",
                 sshkey_user.as_ref().unwrap()
             ))?;
-            let sshkey_keys_v: Vec<String> = sshkey_keys.unwrap().split(" ").map(|s| s.to_string()).collect();
+            let sshkey_keys_v: Vec<String> = sshkey_keys
+                .unwrap()
+                .split(' ')
+                .map(|s| s.to_string())
+                .collect();
             for key in sshkey_keys_v {
                 let key_s: String = key;
-                install_ssh_key(sshkey_user.as_ref().unwrap(), key_s.as_str()).context("Error installing SSH key")?;
+                install_ssh_key(sshkey_user.as_ref().unwrap(), key_s.as_str())
+                    .context("Error installing SSH key")?;
                 log::info!("Installed sshkey: {key_s}");
             }
         }
@@ -821,6 +828,8 @@ mod test {
 
     use super::BinaryFileInProgress;
 
+    use crate::serviceinfo::*;
+
     #[test]
     fn test_binaryfileinprogress_destination_path() {
         assert_eq!(
@@ -835,5 +844,50 @@ mod test {
             BinaryFileInProgress::destination_path(&String::from("/etc/something"), None).unwrap(),
             PathBuf::from("/etc/something")
         );
+    }
+
+    #[test]
+    fn test_pw_encryption() {
+        let type_5_encryption = "$5$ML4hMHtER3/SY9D2$2eWHscoFbfVebDC32qA2dPo3pD6FFM6CRTrvAOMpwQ";
+        assert!(is_password_encrypted(type_5_encryption));
+        let type_2b_encryption = "$2b$ML4hMHtER3/SY9D2$2eWHscoFbfVebDC32qA2dPo3pD6FFM6CRTrvAOMpwQ";
+        assert!(is_password_encrypted(type_2b_encryption));
+        let type_6_encryption = "$6$ML4hMHtER3/SY9D2$2eWHscoFbfVebDC32qA2dPo3pD6FFM6CRTrvAOMpwQ";
+        assert!(is_password_encrypted(type_6_encryption));
+        let plaintext_encryption = "testpassword";
+        assert!(!is_password_encrypted(plaintext_encryption));
+        let empty_pw = "";
+        assert!(!is_password_encrypted(empty_pw));
+    }
+
+    #[test]
+    fn test_user_creation_no_pw() {
+        let test_user = "test";
+        assert!(create_user(test_user).is_ok());
+        let empty_user = "";
+        assert!(create_user(empty_user).is_err());
+        let at_user = "test@test";
+        assert!(create_user(at_user).is_err());
+        let dash_user = "-test";
+        assert!(create_user(dash_user).is_err());
+        let digits_user = "12345";
+        assert!(create_user(digits_user).is_err());
+    }
+
+    #[test]
+    fn test_user_creation_with_pw() {
+        let test_user = "testb";
+        let test_password = "password";
+        assert!(create_user_with_password(test_user, test_password).is_ok());
+        let empty_user = "";
+        assert!(create_user_with_password(empty_user, test_password).is_err());
+        let at_user = "testb@testb";
+        assert!(create_user_with_password(at_user, test_password).is_err());
+        let dash_user = "-testb";
+        assert!(create_user_with_password(dash_user, test_password).is_err());
+        let digits_user = "123456";
+        assert!(create_user_with_password(digits_user, test_password).is_err());
+        let empty_password = "";
+        assert!(create_user_with_password(test_user, empty_password).is_ok());
     }
 }
--- a/client-linuxapp/src/serviceinfo.rs
+++ b/client-linuxapp/src/serviceinfo.rs
@@ -4,6 +4,7 @@ use std::{
     fs::{File, Permissions},
     io::Write,
     path::Path,
+    str,
 };
 use std::{env, fs};
 use std::{os::unix::fs::PermissionsExt, path::PathBuf};
@@ -19,6 +20,8 @@ use fdo_data_formats::{
     types::{CborSimpleTypeExt, Hash, ServiceInfo},
 };
 use fdo_http_wrapper::client::{RequestResult, ServiceClient};
+
+use sha_crypt::{sha256_check, sha256_simple, Sha256Params};
 
 const MAX_SERVICE_INFO_LOOPS: u32 = 1000;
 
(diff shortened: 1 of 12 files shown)


miabbott:
Could you use a Rust crate to handle the password hashing instead of shelling out to `openssl`?

https://docs.rs/password-hash/latest/password_hash/

https://docs.rs/bcrypt/latest/bcrypt/

miabbott:
Also we should be defaulting to SHA512 for the algorithm; MD5 is considered insecure these days.

https://access.redhat.com/articles/880733

client-linuxapp/src/serviceinfo.rs
```diff
@@ -69,6 +72,61 @@ fn create_user(user: &str) -> Result<()> {
...
+    // Checks if user already present
+    let user_info = passwd::Passwd::from_name(user);
+    if user_info.is_some() {
+        log::info!("User {} is already present", user);
+        return Ok(());
+    }
+    // Creates new user if user not present
+    log::info!("Encrypting password");
+    let pw_create = Command::new("echo")
+        .arg(format!("{}|openssl", password))
```
7flying, line 85: In order to pipe one command output to another you need to use `Stdio::piped()` in the source and `.stdin(Stdio::from(x))` in the 'reader', but you need to do it in two steps, first the `echo` part of the command, where you are going to say that `stdout` is going to be piped; and then the `openssl` part as another command, where you are going to say that `stdin` comes from the first command:

```
let echo = Command::new("echo")
        .arg("my password")
        .stdout(Stdio::piped())
        .spawn()
        .expect("some error");
let openssl = Command::new("openssl")
        .arg("passwd")
        .arg("-1")
        .arg("-stdin")
        .stdin(Stdio::from(echo.stdout.expect("another error")))
        .output()
        .expect("yet another error");
```

nullr0ute:
> Could you use a Rust crate to handle the password hashing instead of shelling out to `openssl`?

Or using a function in the openssl crate.

client-linuxapp/src/serviceinfo.rs
```diff
@@ -69,6 +72,61 @@ fn create_user(user: &str) -> Result<()> {
...
+        .arg("-stdin")
+        .stdout(Stdio::piped())
+        .output()
+        .unwrap();
+    let encrypted_pw = String::from_utf8(pw_create.stdout).unwrap();
+    log::info!("Creating user {} with password", user);
+    log::info!("{}", encrypted_pw);
+    Command::new("useradd")
+        .arg("-p")
+        .arg(encrypted_pw)
```
7flying, line 99: you'll possibly need to trim the newline of the `encrypted_pw`

client-linuxapp/src/serviceinfo.rs
```diff
@@ -615,18 +687,30 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
         }
     }
 
-    // Do SSH
+    // Perform SSH or password setup
     if active_modules.contains(&FedoraIotServiceInfoModule::SSHKey.into()) {
-        log::debug!("SSHkey module was active, installing SSH key");
-        if sshkey_user.is_none() || sshkey_key.is_none() {
-            bail!("SSHkey module missing username or key");
+        if sshkey_user.is_none() {
```
7flying, line 692: To make this more readable and less complicated you can make a first `if` with the failure conditions, `bail` there, and then continue handling the rest of the possible options

``` 
if sshkey_user.is_none() || (sshkey.is_some() && sshkey_password.is_some()) {
   bail!();
}
// then we continue with the rest
if something else { 
}
```
but, I think that the story was about allowing to add password and/or sshkey for a user, so both things at the same time should also be allowed.

sarmahaj:
are we not contradicting to FIDO policy of "password less" logins by providing a way to set password for initial user ?
I think we should prefer sshkey rather than password for at least initial user which is for first boot/onboarding. As I recall we did this because for local user it asks for password .

say-paul:
An general thought on having the plain text password in .yaml which may raise some eyebrows.We can force user to put the encrypted password in the file as its done with the `useradd` command , advantage is 1.  simplify the implementation , 2. security.

say-paul:
> are we not contradicting to FIDO policy of "password less" logins by providing a way to set password for initial user ? I think we should prefer sshkey rather than password for at least initial user which is for first boot/onboarding. As I recall we did this because for local user it asks for password .

the ssh_key feature already exists, this added feature for any user who require password to be set also.

client-linuxapp/src/serviceinfo.rs
```diff
@@ -615,18 +687,30 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
...
+            bail!("SSHkey module missing username");
+        } else if sshkey_key.is_some() && sshkey_password.is_some() {
+            bail!("SSHkey module contains both password and key");
```
sarmahaj, line 695: If both (sshkey and password) are present add both .

client-linuxapp/src/serviceinfo.rs
djach7: I think I understand, I'll try that out then. Going to attempt to use a rust or openssl crate first, that sounds like it might be a cleaner implementation.

client-linuxapp/src/serviceinfo.rs
djach7: You're right, I miswrote that. I didn't think about just having all the bails together, thank you for the idea

djach7:
Thank you all for the comments/ideas, I'm going to try to understand the openssl crate and I'll make the other changes above too.

client-linuxapp/src/serviceinfo.rs
```diff
@@ -69,6 +74,54 @@ fn create_user(user: &str) -> Result<()> {
...
+    // Checks if user already present
+    let user_info = passwd::Passwd::from_name(user);
+    if user_info.is_some() {
+        log::info!("User {} is already present", user);
+        return Ok(());
+    }
+    
+    log::info!("Encrypting password");
+    let default_params: Sha256Params = Default::default();
+    let str_encrypted_pw = sha256_simple(password, &default_params).expect("Hashing failed");
```
7flying, line 87: don't forget the `?` after the `expect`, we want to propagate the errors to the caller function

client-linuxapp/src/serviceinfo.rs
```diff
@@ -69,6 +74,54 @@ fn create_user(user: &str) -> Result<()> {
     Ok(())
 }
 
+fn create_user_with_password(user: &str, password: &str) -> Result<()> {
+    // Checks if user already present
+    let user_info = passwd::Passwd::from_name(user);
+    if user_info.is_some() {
+        log::info!("User {} is already present", user);
```
7flying, line 81: This is just a nit: rust allows to do `log::info!("{user}");` (and with the `format!` macro, `println!` and some others) which in this case is equivalent to `log::info!("{}", user);`, but it is considered to be more clean, if you can use inline format parameters, you are encouraged to use them (see [this](https://doc.rust-lang.org/stable/std/fmt/)).

Note that it doesn't work if you need to do something complex with the variable before:
```
// this does not compile:
format!("Error creating new user with password: {sshkey_user.as_ref().unwrap()}");
```

djach7: Thank you for sharing! I just updated my code to use the inline format where I could

client-linuxapp/src/serviceinfo.rs
djach7: I tried to add in the `?` after `.expect("Hashing failed")` but I was given the error below. I can put more time into it once I get ssh working again. 
```
error[E0277]: the `?` operator can only be applied to values that implement `Try`
  --> client-linuxapp/src/serviceinfo.rs:87:28
   |
87 |     let str_encrypted_pw = sha256_simple(password, &default_params).expect("Hashing failed")?;
   |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the `?` operator cannot be applied to type `String`
   |
   = help: the trait `Try` is not implemented for `String`
   ```

Comment (7flying) [reply client-linuxapp/src/serviceinfo.rs:87]: I see :thinking: , all good then

File: client-linuxapp/src/serviceinfo.rs [hunk 7]
```diff
@@ -74,22 +74,37 @@ fn create_user(user: &str) -> Result<()> {
     Ok(())
 }
 
+fn check_password_encryption(s: &str) -> bool {
+    let prefixes = ["$2b$", "$6$", "$5$"];
+
+    for prefix in prefixes.iter() {
```
7flying, line 80: there is no need for the `.iter()` there

client-linuxapp/src/serviceinfo.rs
```diff
@@ -74,22 +74,37 @@ fn create_user(user: &str) -> Result<()> {
...
+        if s.starts_with(prefix) {
+            return true;
+        }
+    }
+
+    return false;
```
7flying, line 86: no need for the `return false;` there, simply `false` and the expression will be returned

client-linuxapp/src/serviceinfo.rs
```diff
@@ -74,22 +74,37 @@ fn create_user(user: &str) -> Result<()> {
     Ok(())
 }
 
+fn check_password_encryption(s: &str) -> bool {
```
7flying, line 77: it is nice to add a comment to the original source if you have been inspired by some other source code for this function

client-linuxapp/src/serviceinfo.rs
```diff
@@ -69,6 +74,49 @@ fn create_user(user: &str) -> Result<()> {
...
+    // Checks if user already present
+    let user_info = passwd::Passwd::from_name(user);
+    if user_info.is_some() {
+        log::info!("User {user} is already present");
+        return Ok(());
+    }
+
+    let mut str_encrypted_pw = password.to_string();
+    log::info!("Checking for password encryption");
+    if check_password_encryption(password) == false {
```
7flying, line 99: It would make the goal of the function more clear if you rename it to `is_password_encrypted`, for instance.

Don't compare results to true or false, is best to: `if !is_password_encrypted(password) {`

client-linuxapp/src/serviceinfo.rs
```diff
@@ -615,18 +677,34 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
...
+                install_ssh_key(sshkey_user.as_ref().unwrap(), key_s.as_str()).context("Error installing SSH key")?;
+                log::info!("Installed sshkey: {key_s}");
+            }
+        }
+        if sshkey_password.is_some() {
+            log::info!("SSHkey module was active, creating user with password");
+            create_user_with_password(sshkey_user.as_ref().unwrap(), sshkey_password.as_ref().unwrap()).context(format!(
+                "Error creating new user with password: {}",
+                sshkey_user.as_ref().unwrap() 
+            ))?;
```
7flying, line 706: I might be wrong, but I think that we do want to add both ssh key/s and a password for a given user if that is given to us.
With this as it is, the `create_user` and  `create_user_with_password` functions both return when a user is already created, so it is not possible.

client-linuxapp/src/serviceinfo.rs
```diff
@@ -67,9 +69,58 @@ fn create_user(user: &str) -> Result<()> {
...
         .wait()
         .context("Error creating new user")?;
+    // Command::new("useradd")
+    //     .arg("-p")
+    //     .arg(format!("$(openssl passwd -1 {})", password))
+    //     .arg(user)
+    //     .spawn()
+    //     .context("Error spawning new user command")?
+    //     .wait()
+    //     .context("Error creating new user")?;
```
miabbott: nit: can clean up these unused comments

client-linuxapp/src/serviceinfo.rs
```diff
@@ -67,9 +69,58 @@ fn create_user(user: &str) -> Result<()> {
...
+    //     .spawn()
+    //     .context("Error spawning first command to enter password")?
+    //     .wait()
+    //     .context("Error entering password first time")?;
+    // // Handles system's second prompt to type new password
+    // Command::new(password)
+    //     .spawn()
+    //     .context("Error spawning second command to enter password")?
+    //     .wait()
+    //     .context("Error entering password second time")?;
```
miabbott: nit: can remove these unused comments

client-linuxapp/src/serviceinfo.rs
```diff
@@ -405,11 +457,27 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
             bail!("Non-activated module {} got request", module);
         }
         if module == FedoraIotServiceInfoModule::SSHKey.into() {
-            let value = value.as_str().context("Error parsing sshkey value")?;
+            // let value = value
+            //     .as_str()
+            //     .context("Error parsing sshkey value")?;
```
miabbott: nit: can clean up these unused comments

serviceinfo-api-server/src/main.rs
```diff
@@ -251,6 +252,15 @@ async fn serviceinfo_handler(
...
         };
+=======
+        if let Some(initial_user) = &user_data.service_info_configuration.settings.initial_user {
+            reply.reply.initial_user = Some(ServiceInfoApiReplyInitialUser {
+                username: initial_user.username.clone(),
+                password: initial_user.password.clone(),
+                ssh_keys: initial_user.sshkeys.clone(),
+            });
+        }
+>>>>>>> dcea215 (Building blocks for pw implementation, very broken)
```
miabbott: Looks like some leftover cruft from a rebase?

owner-onboarding-server/src/handlers.rs
```diff
@@ -538,9 +538,15 @@ async fn perform_service_info(
...
             &initial_user.username,
         )?;
+        out_si.add(
+            FedoraIotServiceInfoModule::SSHKey, 
+            "password", 
+            &initial_user.password,
+        )?;
         for key in initial_user.ssh_keys.iter() {
-            out_si.add(FedoraIotServiceInfoModule::SSHKey, "key", &key)?;
+             out_si.add(FedoraIotServiceInfoModule::SSHKey, "key", &key)?;
```
miabbott: nit: extra leading space

miabbott:
Oh!  I was commenting on the individual commits and it looks like the cleanups I suggested were done in a later revision.

It might be nice to clean up that first commit so there's nothing dirty in the commit history to allow for cleaner bisection/revert

client-linuxapp/src/serviceinfo.rs
```diff
@@ -70,6 +75,48 @@ fn create_user(user: &str) -> Result<()> {
     Ok(())
 }
 
+// Returns true if password is encrypted
+// is_password_encrypted functionality is taken from osbuild-composer's crypt.go
```
miabbott, line 79: nit: might be nice to include a link to the implementation in crypt.go

djach7: Resolved, link added

nullr0ute:
> Oh! I was commenting on the individual commits and it looks like the cleanups I suggested were done in a later revision.
> 
> It might be nice to clean up that first commit so there's nothing dirty in the commit history to allow for cleaner bisection/revert

It's likely it should be compacted to a single commit, or if multiple commits each commit should do something specific.

djach7:
Just squashed them together since that does make more sense. Looks like some rebase stuff still exists in the first commit that doesn't need to be there so I'll figure out how to get rid of that. Going to resolve all of the outdated comments too.

client-linuxapp/src/serviceinfo.rs
```diff
@@ -617,18 +673,42 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
...
+                .unwrap()
+                .split(' ')
+                .map(|s| s.to_string())
+                .collect();
+            for key in sshkey_keys_v {
+                let key_s: String = key;
+                install_ssh_key(sshkey_user.as_ref().unwrap(), key_s.as_str())
+                    .context("Error installing SSH key")?;
+                log::info!("Installed sshkey: {key_s}");
+            }
```
7flying, line 710: One of the things that's different from before is how we handle multiple ssh keys, can you modify the e2e tests so that it installs two ssh keys?

nullr0ute: If two keys is new functionality in this PR it should be a separate commit with the updated tests in that commit.

miabbott: In the realm of multiple SSH keys, it may be useful to consider `ssh-key-dir` in the future - https://github.com/coreos/ssh-key-dir

miabbott:
If I'm reading the code correctly (and I may not be!), it looks like we only have a single test for adding a plaintext password as part of onboarding.

Should we expand the test cases to at least cover a crypted/hashed password?  Maybe some other negative test cases (i.e. null passwords, unsupported hashes, missing username, etc)?

djach7:
I can add more tests if we'd like, I was following the example from the sshkeys where as far as I could tell there was just one primary test for having one sshkey (but maybe I'm just reading the code wrong?)

client-linuxapp/src/serviceinfo.rs
```diff
@@ -617,18 +681,33 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
...
-    // Do SSH
+    // Perform SSH or password setup
     if active_modules.contains(&FedoraIotServiceInfoModule::SSHKey.into()) {
-        log::debug!("SSHkey module was active, installing SSH key");
-        if sshkey_user.is_none() || sshkey_key.is_none() {
-            bail!("SSHkey module missing username or key");
+        if sshkey_user.is_none() {
+            bail!("SSHkey module missing username");
+        } else if sshkey_keys.is_none() && sshkey_password.is_none() {
+            bail!("SSHkey module missing password and key")
```
7flying: just for consistency, add a `;`  at the end of this statement

7flying:
There are some changes made into the `serviceinfo-api-server`, `owner-onboarding-server` and `client-linuxapp` on the  `chore(integration-tests)` commit that are not tests, I would like those changes to be moved to the implementation commit.

Other than that, we are not testing that you can add more than one ssh key per user, but we can add that as a follow-up PR.

So, feel free to mark the PR as ready

nullr0ute:
> Other than that, we are not testing that you can add more than one ssh key per user, but we can add that as a follow-up PR.

Can we ensure an issue is created so that isn't lost.

7flying:
> > Other than that, we are not testing that you can add more than one ssh key per user, but we can add that as a follow-up PR.
> 
> Can we ensure an issue is created so that isn't lost.

Yep, linking the issue here for future reference #507

client-linuxapp/src/serviceinfo.rs
```diff
@@ -54,22 +57,82 @@ fn set_perm_mode(path: &Path, mode: u32) -> Result<()> {
...
+        .spawn()
+        .context("Error spawning new user command")?
+        .wait()
+        .context("Error creating new user")?;
+
+    if status.success() {
+        log::info!("User {user} created successfully with password");
+        Ok(())
+    } else {
+        bail!(format!("User creation failed. Exit Status: {status}"));
```
7flying, line 130: you need `status.code()` to get the internal status code

7flying: there is another instance of this above

7flying (approved):
Everything looks good, thank you!

7flying (approved):

Issue #403, opened 2023-01-27
Cannot disable serviceinfo auth


waldner:
Despite the docs saying that `service_info_auth_token` can be set to `None` in `serviceinfo_api_server.yml` if no authentication is needed, it seems that the server is always looking for authentication headers even with `service_info_auth_token: None`, as seen in the following log:

```
 2023-01-27T17:27:53.946Z INFO  fdo_serviceinfo_api_server > Listening on 0.0.0.0:8083
 2023-01-27T17:29:32.821Z DEBUG hyper::proto::h1::io       > parsed 2 headers
 2023-01-27T17:29:32.821Z DEBUG hyper::proto::h1::conn     > incoming body is empty
 2023-01-27T17:29:32.821Z INFO  serviceinfo-api-server     > 127.0.0.1:48710 "GET /device_info HTTP/1.1" 400 "-" "-" 80.245µs
 2023-01-27T17:29:32.821Z DEBUG warp::filter::service      > rejected: Rejection([MethodNotAllowed, MethodNotAllowed, MissingHeader { name: "Authorization" }])
 2023-01-27T17:29:32.821Z DEBUG hyper::proto::h1::io       > flushed 164 bytes
 2023-01-27T17:31:02.826Z DEBUG hyper::proto::h1::conn     > read eof
...
```

7flying:
Can reproduce on `main`, will check if it's a bug or the docs are outdated and we always need authentication.

PR #441, opened 2023-03-09
fix(serviceinfo): no authentication is needed when omitting `service_info_auth_token`

nak3:
This patch supports no authentication for service info by using `Option()`
for `service_info_auth_token`.

Current docs asks users to set `None` if no authentication is needed but
`serde_yaml` library does not handle the `None` as Rust's `None`. 
So this patch supports `service_info_auth_token` as `Option()` to allow
"no authentication is needed" when it is omitted.

Alternatively it is possible to support `"None"` or `""` (empty string)
as a special case to disable authorization. Any suggestion is welcome.

Fix https://github.com/fedora-iot/fido-device-onboard-rs/issues/403

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -469,8 +469,8 @@ service_info:
 
 Where:
 - `bind`: IP address and port that the Service Info API Server will take.
-- `service_info_auth_token`: Authorization token, `None` if no authentication
-  is needed.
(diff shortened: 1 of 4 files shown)


nak3:
Thank you for the review! Sure I have updated it.

7flying (approved):
LGTM!

PR #446, opened 2023-03-10
[DO NOT MERGE] just a test to validate something...

runcom:


runcom:
@7flying @sarmahaj take a look here, if we get the error from Juerg, then we know what this is :)

7flying:
Closing since we already have tests for this

PR #451, opened 2023-03-27
Allow MAC addresses as mfg_string_type

### LINKED ISSUE #447: Using `StructuredDeviceInfo` as the `mfg_string_type` may break the X509 certificate standard's max length
When using the `StructuredDeviceInfo` as the `mfg_string_type` we might get a string that is longer than what the X509 certificate standard allows in its `CN` field, this causes an error in the Manufacturing Server, at the openssl `append_entry_by_text` call on `create_device_certificate`:

```
2023-03-10T13:11:14.902Z ERROR fdo_http_wrapper::server                 > Error occurred while processing message type DIAppStart: ErrorStack([Error { code: 109052055, library: "asn1 encoding routines", function: "ASN1_mbstring_ncopy", reason: "string too long", file: "crypto/asn1/a_mbstr.c", line: 106, data: "maxsize=64" }])
```

The max size of this field is: ` assert!(value.len() <= c_int::max_value() as usize);`

#### To take into account
- input fields that end up in the Manufacturing Server such as the serial/structured-device-info should be lenght-checked in the client if possible rather than wait until they are passed to the Manufacturing Server

7flying:
Adds the `MfgStringType::MACAddress` to the device identification
methods. This MACAddress will be selected from an iface specified
by the environment variable `DI_MFG_STRING_TYPE_MAC_IFACE` once the
env var `DI_MFG_STRING_TYPE` has been set to `mac_address`.

This commit also allows to call the manufacturing-client with a
command-line interface as an alternative to the env variable method.

Closes #447


manufacturing-client/src/main.rs
```diff
@@ -240,130 +309,154 @@ async fn main() -> Result<()> {
...
-                    if non_zero_mac {
-                        write!(serial_mac, "IFACE_{interface}={mac};").unwrap();
-                        log::debug!("MAC address {serial_mac} of interface {interface} ");
-                    }
-                }
+        MfgStringType::MACAddress => {
+            if iface.is_none() {
+                bail!("No iface provided");
+            }
+            let given_iface = iface.unwrap();
```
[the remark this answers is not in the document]
7flying: The unwraps that I've put are on a helper function of the unit tests, were I don't care that it won't error gracefully honestly, I just want it to error as soon as possible. I will add `context` regardless :+1:

7flying:
> cc @jritter

you would use this like:
```
export MANUFACTURING_SERVER_URL=http://0.0.0.0:8080
export DIUN_PUB_KEY_INSECURE="true"
export LOG_LEVEL=trace
export DI_MFG_STRING_TYPE=mac_address
export DI_MFG_STRING_TYPE_MAC_IFACE=enp1s0
./target/debug/fdo-manufacturing-client
```
manufacturing server config (just relevant changes):
```yaml
protocols:
  plain_di: false
  diun:
    mfg_string_type: MACAddress
    key_type: SECP384R1
    allowed_key_storage_types:
    - FileSystem
    key_path: /home/fedora/fido-device-onboard-rs/aio-dir/keys/diun_key.der
    cert_path: /home/fedora/fido-device-onboard-rs/aio-dir/keys/diun_cert.pem
```

manufacturing-client/src/main.rs
```diff
@@ -240,130 +309,164 @@ async fn main() -> Result<()> {
...
-                            break;
-                        }
-                    }
-                    if non_zero_mac {
-                        write!(serial_mac, "IFACE_{interface}={mac};").unwrap();
-                        log::debug!("MAC address {serial_mac} of interface {interface} ");
-                    }
-                }
+        MfgStringType::MACAddress => {
+            let given_iface = iface.context("No iface provided")?;
```
sarmahaj, line 444: the check on iface is done already in main() , so this check is optional ,I guess.

integration-tests/tests/di_diun.rs
```diff
@@ -71,3 +75,91 @@ async fn test_device_credentials_already_active() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_device_credentials_generated_with_mac_address() -> Result<()> {
```
sarmahaj, line 80: Here can we add a test inside by running manufacturing client separately as `plain DI` and `DIUN` , so that we cover both cases

miabbott:
Could we add some docs on how to use the new device identification features?

We probably should have done this as part of #396, so it's not specific to this PR.

Perhaps a new issue to track additional docs would suffice for my request at this time.

manufacturing-client/src/main.rs
7flying: yeah, but we should never use un-checked inputs on functions since the function might be called from somewhere else other than main in the future

runcom:
and here we have the c9s build https://dashboard.packit.dev/results/copr-builds/686651

runcom:
one thing we need is to get this to main too - as we need to build and upload latest containers too (we don't publish containers for every PR

runcom (approved):

PR #457, opened 2023-03-29
feat: integrate Packit with COPR

### LINKED ISSUE #435: Setup Packit for the repo
It would be excellent if we could configure [Packit](https://packit.dev/) on this repo, so that we could simplify the RPM packaging process.

I don't know if there are additional challenges around packaging Rust code + dependent crates in Fedora/CentOS Stream, but I hope we wouldn't be the first to take on that challenge.

(Having Packit configured would also help for requests like this: https://github.com/fedora-iot/fido-device-onboard-rs/issues/333#issuecomment-1450537680)

runcom:
fix #435

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -35,3 +35,13 @@ jobs:
   - centos-stream-9-x86_64
   - fedora-development-aarch64
   - fedora-development
+- job: copr_build
+  trigger: commit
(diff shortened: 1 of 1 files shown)


7flying (approved):
LGTM!!!

Issue #463, opened 2023-04-01
assertion for the encode item in cborparser is not accurate


nak3:
The assertion has `var <= 24`

https://github.com/fedora-iot/fido-device-onboard-rs/blob/874b97e3236e6a3c9e57fd9ea5f4d177e3a80ab6/data-formats/src/cborparser.rs#L153

but `24` does not fit a single byte (it will be `[MAJOR_TYPE|24, val]`) so the assertion should be `assert!(val < 24);`.

(With said, `encode_item_start` should support the value larger than 23 like `read_len()` does.)

nak3:
Current code fails with the test in https://github.com/fedora-iot/fido-device-onboard-rs/pull/464 with:

```
---- cborparser::test::test_value_with_large_value stdout ----
thread 'cborparser::test::test_value_with_large_value' panicked at 'assertion failed: `(left == right)`
  left: `[216, 129, 1]`,
 right: `[216, 24, 129, 1]`', data-formats/src/cborparser.rs:790:9


failures:
    cborparser::test::test_value_with_large_value
```

It is not failed by assertion but the mismatch of the serialization.

PR #464, opened 2023-04-01
feat(data-formats): Support encoding the large number by cborparser

nak3:
As per `TODO` comment, current code does not support the value larther than 23 even though `read_len()` function supports it.

This patch adds the support.

Fix https://github.com/fedora-iot/fido-device-onboard-rs/issues/463


7flying (approved):
Sorry for the delay, had to try to make sense of the cbor standard.
LGTM!

PR #465, opened 2023-04-01
chore: remove unused function `get_hash_two_items`

nak3:
This patch removes drops `get_hash_two_items` which is not called anywhere after the commit 865ca08.


7flying (approved):
LGTM

PR #476, opened 2023-04-14
feat(fdo-manufacturing-client): Support command line option for DI_SIGN_KEY_PATH and DI_HMAC_KEY_PATH

nak3:
As `fdo-manufacturing-client plain-di` does not have the option for `DI_SIGN_KEY_PATH` and `DI_HMAC_KEY_PATH`, we need to specify them via env variable.

This patch support the command line option.

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -679,9 +679,15 @@ identification with the Manufacturing server.
    Identification keys, valid values: `filesystem` (`tpm` not yet
    implemented).
    
-3. `DI_SIGN_KEY_PATH`:
+3. `DI_SIGN_KEY_PATH`: filepath to the private key with DER format.
+   For example, you can create the key file with `openssl` command:
+   ```sh
+   $ openssl ecparam -name secp384r1 -genkey -noout -outform der -out /path/to/sign_key/key.der
+   ```
+   Please note that the curve `secp384r1` must match to the `key_type` specified in `manufacturing_server.yaml`.
 
-4. `DI_HMAC_KEY_PATH`:
+4. `DI_HMAC_KEY_PATH`: filepath to the HMAC key. There is no format limitation,
+   you can use the file with random value.
 
 5. `DEVICE_CREDENTIAL_FILENAME`: [optional] filepath specified by the user to
    store the device credentials, by default `/etc/device-credentials`.


7flying:
I did not add those options since I'm usure of the values that they should take. Can you provide an example?

nak3:
Sorry if you already might know but these values (keys) are stored in device credential. You can follow the code `perform_di()` -> `save_to_credential()`.

The device credential stores the keys at last as:

https://github.com/fedora-iot/fido-device-onboard-rs/blob/69fec227de04d98c51a5b5d8eb218bdc264f0a3f/data-formats/src/devicecredential/file.rs#L24-L27

You can dump the secrets by `fdo-owner-tool dump-device-credential`, though you need to tweak the code to print the secret.

7flying:
> Sorry if you already might know but these values (keys) are stored in device credential. You can follow the code `perform_di()` -> `save_to_credential()`.
> 
> The device credential stores the keys at last as:
> 
> https://github.com/fedora-iot/fido-device-onboard-rs/blob/69fec227de04d98c51a5b5d8eb218bdc264f0a3f/data-formats/src/devicecredential/file.rs#L24-L27
> 
> You can dump the secrets by `fdo-owner-tool dump-device-credential`, though you need to tweak the code to print the secret.

Yes, my question is about how a user would generate those secrets. What type of keys are they? How do I generate them?

nak3:
Oh, I see. Sorry for my misreading. I hope below answers your question.

#### hmac_key_path:
* It is a plain file contains a random value. (if you don't concern about security, any value is fine and there is no restrictions for 32 byte.)
* Example command to generate:
```
$ head --bytes 32 /dev/urandom  > /path/to/hmac/secret
```
#### sign_key_path

* The private key which is a DER format.
* Example command to generate:
```
$ openssl ecparam -name secp384r1 -genkey -noout -outform der -out /path/to/sign_key/key.der
```

7flying:
I'm just going to see what we can do for #477 before I merge this.
We need some documentation/guidelines on how to generate the keys and were to put them before this goes in.

nak3:
Alright, I updated the docs https://github.com/nak3/fido-device-onboard-rs/blob/add-key-path-option/HOWTO.md#plain-di 
Is it alright? Or do we need a doc somewhere else?

PR #479, opened 2023-04-18
feat(serviceinfo_api_server): implement per device Serviceinfo initial user and sshkey

sarmahaj:
This PR 
- provides a way to user for adding different initial user and sshkey by using `device_specific_store_driver` feature.
- user has to provide a .yml file with the same name GUID of the device under `device_specific_store_driver` path.
- If no file is present for a device then base config is applied (from serviceinfo_api_server.yml).
- updated HOWTO.md

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -22,6 +22,8 @@
 - How to run the clients:
   - Linuxapp client
   - Manufacturing client
+- How to use Features:
+  - How to use the `per-device serviceinfo` feature
 
 ## Pre-requisites
 
@@ -734,3 +736,26 @@ Options:
 
 Please note that in this mode there are some environment variables that are
 still required to be set by the user (`DI_SIGN_KEY_PATH`, `DI_HMAC_KEY_PATH`).
+
+## How to use Features
+
+### How to use the `per-device serviceinfo` feature
+
+  Using this feature the user can choose to apply different serviceinfo settings on different devices.
+  For that the user needs to provide a path to a `per-device serviceinfo` file under the `device_specific_store_driver` field
+  present in the `serviceinfo_api_server.yml` file.
+  If other devices do not have their `per-device serviceinfo` file under `device_specific_store_driver` they will get onboarded
+  with settings from the main file, which is `serviceinfo_api_server.yml`.
+  
+  1. Initialize the device as mentioned in [How to generate an Ownership Voucher and Credential for a Device](#how-to-generate-an-ownership-voucher-ov-and-credential-for-a-device-device-initialization).
+
+  2. Dump the `device-credentials`
+  ```bash
+  fdo-owner-tool dump-device-credential /path/to/device-credentials
+  ```
+
+  3. Note the GUID of the device and create a .yml file with same name as the `guid` under directory path `device_specific_store_driver`.
+
+  4. You can refer to [per_device_serviceinfo.yml](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/examples/config/device_specific_serviceinfo.yml) as an example.
+
+  5. Follow the onboarding procedure and this particular device will get the serviceinfo settings as mentioned in the above file.
--- a/integration-tests/tests/common/mod.rs
+++ b/integration-tests/tests/common/mod.rs
@@ -778,6 +778,16 @@ impl<'a> TestServerConfigurator<'a> {
             .runner_path(&self.server_number)
             .join(config_file_name);
 
+        let per_device: bool = match env::var("PER_DEVICE_SERVICEINFO") {
+            Ok(val) => val.parse().unwrap_or(false),
+            Err(e) => {
+                eprintln!(
+                    "Error reading environment variable: {} setting to default",
+                    e
+                );
+                false
+            }
+        };
         self.test_context
             .generate_config_file(&output_path, config_file_name, |cfg| {
                 cfg.insert(
@@ -790,10 +800,23 @@ impl<'a> TestServerConfigurator<'a> {
                     "config_dir",
                     &self.test_context.runner_path(&self.server_number),
                 );
-                cfg.insert(
-                    "user",
-                    users::get_current_username().unwrap().to_str().unwrap(),
-                );
+
+                if !per_device {
+                    L.l("per_device_serviceinfo is not set, using default values");
+                    cfg.insert(
+                        "user",
+                        users::get_current_username().unwrap().to_str().unwrap(),
+                    );
+                    cfg.insert("sshkey", "sshkey_default");
+                } else {
+                    L.l("per_device_serviceinfo is set, using device specific values");
+                    cfg.insert(
+                        "user",
+                        users::get_current_username().unwrap().to_str().unwrap(),
+                    );
+                    cfg.insert("sshkey", "sshkey_per_device");
+                }
+
                 // TODO: Insert more defaults
 
                 context_configurator(cfg)
--- a/integration-tests/tests/e2e.rs
+++ b/integration-tests/tests/e2e.rs
@@ -63,11 +63,38 @@ async fn test_e2e() -> Result<()> {
 
     let mut failed = Vec::new();
 
+    env::set_var("PER_DEVICE_SERVICEINFO", "false");
+
     for (diun_verification_method_name, diun_verification_method_generator) in
         &diun_verification_methods[1..2]
     {
+        // for default serviceinfo
         for diun_key_type in &diun_key_types {
-            match test_e2e_impl(diun_verification_method_generator, diun_key_type).await {
+            match test_e2e_impl_default_serviceinfo(
+                diun_verification_method_generator,
+                diun_key_type,
+            )
+            .await
+            {
+                Ok(()) => (),
+                Err(e) => {
+                    failed.push(TestCase {
+                        diun_verification_method_name: diun_verification_method_name,
+                        diun_key_type: diun_key_type,
+                        error: e,
+                    });
+                }
+            }
+        }
+
+        // for per_device serviceinfo
+        for diun_key_type in &diun_key_types {
+            match test_e2e_impl_per_device_serviceinfo(
+                diun_verification_method_generator,
+                diun_key_type,
+            )
+            .await
+            {
                 Ok(()) => (),
                 Err(e) => {
                     failed.push(TestCase {
@@ -90,15 +117,6 @@ async fn test_e2e() -> Result<()> {
     }
 }
 
-#[derive(Debug, serde::Serialize)]
-struct ServiceInfoApiAdminV0Request {
-    device_guid: String,
-    // This is not entirely accurate (it accepts any JSON value), but for this test sufficient
-    service_info: Vec<(String, String, serde_json::Value)>,
-}
-
-const CI_TESTSTRING: &str = "CI_TESTSTRING_SHOULD_BE_RETURNED_IN_SERVICEINFO";
-
 #[derive(Debug)]
 struct TestCase {
     #[allow(dead_code)]
@@ -109,13 +127,17 @@ struct TestCase {
     error: anyhow::Error,
 }
 
-async fn test_e2e_impl<F>(verification_generator: F, diun_key_type: &str) -> Result<()>
+async fn test_e2e_impl_default_serviceinfo<F>(
+    verification_generator: F,
+    diun_key_type: &str,
+) -> Result<()>
 where
     F: Fn(&TestContext) -> Result<(&'static str, String, &'static str)>,
 {
     let ci = env::var("FDO_PRIVILEGED").is_ok();
+    env::set_var("PER_DEVICE_SERVICEINFO", "false");
     let mut ctx = TestContext::new().context("Error building test context")?;
-    let new_user: &str = "testuser"; //new user to be created during onboarding
+    let new_user: &str = "testuser"; // new user to be created during onboarding
     let encrypted_disk_loc = ctx.testpath().join("encrypted.img");
     let rendezvous_server = ctx
         .start_test_server(
@@ -317,35 +339,12 @@ where
         bail!("Failed to call clevis luks bind");
     }
 
-    #[allow(unused_mut)]
-    let mut service_info = vec![(
-        "CI".to_string(),
-        "teststring".to_string(),
-        serde_json::Value::String(CI_TESTSTRING.to_string()),
-    )];
-
     let client = reqwest::Client::new();
-    // Submit additional ServiceInfo for this device
-    client
-        .post(format!(
-            "http://localhost:{}/admin/v0", //DevSkim: ignore DS137138
-            serviceinfo_api_server.server_port().unwrap()
-        ))
-        .header("Authorization", "Bearer TestAdminToken")
-        .json(&ServiceInfoApiAdminV0Request {
-            device_guid: device_guid,
-            service_info: service_info,
-        })
-        .send()
-        .await
-        .context("Error sending ServiceInfo API request")?
-        .error_for_status()
-        .context("Error from the ServiceInfo API request")?;
 
     // Ensure TO0 is executed
     let res = client
         .post(format!(
-            "http://localhost:{}/report-to-rendezvous", //DevSkim: ignore DS137138
+            "http://localhost:{}/report-to-rendezvous", // DevSkim: ignore DS137138
             owner_onboarding_server.server_port().unwrap()
         ))
         .send()
@@ -381,7 +380,6 @@ where
         )
         .context("Error running client")?;
     output.expect_success().context("client failed")?;
-    output.expect_stderr_line(CI_TESTSTRING)?;
 
     pretty_assertions::assert_eq!(
         fs::read_to_string(&marker_file_path).context("Error reading marker file")?,
@@ -392,7 +390,7 @@ where
             .context("Error reading authorized SSH keys")?,
         "
 # These keys are installed by FIDO Device Onboarding
-testkey
+sshkey_default
 # End of FIDO Device Onboarding keys
 "
     );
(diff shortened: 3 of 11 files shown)


serviceinfo-api-server/src/main.rs
```diff
@@ -224,7 +227,64 @@ async fn serviceinfo_handler(
...
+
+        let path_per_device_store = match settings.device_specific_store_driver {
+            StoreConfig::Directory { mut path } => {
+                let file_name = format!(
+                    "{}.yml",
+                    query_info.device_guid.to_string().replace('\"', "")
+                );
+                path.push(file_name);
+                path.to_string_lossy().into_owned()
+            }
```
runcom, line 248: I'd mask all of this behind the store no? ultimately, we do want to do `device_specific_store.load_data(device_uuid)`

7flying: Agreed

util/src/servers/configuration/serviceinfo_api_server.rs
```diff
@@ -19,6 +19,7 @@ pub struct ServiceInfoApiServerSettings {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(deny_unknown_fields)]
```
7flying, line 22: We can delete that now that we know the correct format

serviceinfo-api-server/src/main.rs
```diff
@@ -1,19 +1,22 @@
...
+    settings_for, settings_per_device, ServiceInfoApiReply, ServiceInfoApiReplyInitialUser,
+    ServiceInfoApiReplyReboot,
 };
+use serde::{Deserialize, Serialize};
+use std::{collections::HashSet, path::Path, str::FromStr};
+use tokio::signal::unix::{signal, SignalKind};
+use warp::Filter;
+
+#[derive(Debug)]
+struct RtrFailure(anyhow::Error);
```
7flying, line 18: We named this struct like that somewhere else in the code because `RtrFailure` was encapsulating a `Return to Rendezvous` failure. In this case we should name the reject something else.

serviceinfo-api-server/src/main.rs
```diff
@@ -224,7 +227,64 @@ async fn serviceinfo_handler(
...
+                path.to_string_lossy().into_owned()
+            }
+        };
+
+        if Path::new(&path_per_device_store).exists() {
+            //device specific config
+            let per_device_settings = match settings_per_device(&path_per_device_store) {
+                Ok(config) => config.try_deserialize::<ServiceInfoSettings>(),
+                Err(e) => {
+                    return Err(warp::reject::custom(RtrFailure(e)));
```
7flying, line 256: Since we are returning an error to the client, if the file does not exist or something similar, we do not want to leak the path were the things are stored in the server, so I would log the error in the server and return a more generic error message to the client.

util/src/servers/configuration/serviceinfo_api_server.rs
sarmahaj: I think its better to add incase there is any wrong field/typo during creating .yml file , this macro will show error that some unknown field in the .yml, easy for debugging.

serviceinfo-api-server/src/main.rs
sarmahaj: Ahh I thought its returnfailure something like that.. will change it to something better

HOWTO.md
```diff
@@ -734,3 +736,31 @@ Options:
 
 Please note that in this mode there are some environment variables that are
 still required to be set by the user (`DI_SIGN_KEY_PATH`, `DI_HMAC_KEY_PATH`).
+
+## How to use Features
+  ### How to use `device_specific_store_driver` feature
```
7flying, line 741: no need to indent the `###` to the right here, and a newline above and below too

HOWTO.md
```diff
@@ -734,3 +736,31 @@ Options:
...
+  - Using this feaure user can choose to apply different serviceinfo settings to different devices.
```
7flying, line 742: Since this is a general explanation of how the feature works I would put it outside the `-` as a general explanation.

typo: "... this feature the user"  "settings on different devices."

HOWTO.md
```diff
@@ -734,3 +736,31 @@ Options:
...
+  For that user needs to provide path to device specific file under `device_specific_store_driver` field
+  present in the `serviceinfo_api_server.yml` file.
+  - File name should be name `guid` of the device.
```
7flying, line 745: typo: "The file name should be the `guid`..."

HOWTO.md
```diff
@@ -734,3 +736,31 @@ Options:
...
+  - Initialize the device as mentioned in [How to generate an Ownership Voucher and Credential for a Device](#how-to-generate-an-ownership-voucher-ov-and-credential-for-a-device-device-initialization).
+  - Dump the `device-credentials`
+    ```bash
+    fdo-owner-tool dump-device-credential /path/to/device-credentials
+    ```
+  - Note the GUID of the device and create a .yml with same name under directory path `device_specific_store_driver`.
```
7flying, line 751: here you are repeating the GUID part that you have mentioned above, we can try to summarise it

HOWTO.md
```diff
@@ -734,3 +736,31 @@ Options:
 
 Please note that in this mode there are some environment variables that are
 still required to be set by the user (`DI_SIGN_KEY_PATH`, `DI_HMAC_KEY_PATH`).
+
+## How to use Features
+  ### How to use `device_specific_store_driver` feature
+  - Using this feaure user can choose to apply different serviceinfo settings to different devices.
+  For that user needs to provide path to device specific file under `device_specific_store_driver` field
```
7flying, line 743: typo: "the user needs to provide a path to a device-specific file under the"

HOWTO.md
```diff
@@ -734,3 +736,31 @@ Options:
...
+  present in the `serviceinfo_api_server.yml` file.
+  - File name should be name `guid` of the device.
+  - Initialize the device as mentioned in [How to generate an Ownership Voucher and Credential for a Device](#how-to-generate-an-ownership-voucher-ov-and-credential-for-a-device-device-initialization).
+  - Dump the `device-credentials`
+    ```bash
+    fdo-owner-tool dump-device-credential /path/to/device-credentials
+    ```
+  - Note the GUID of the device and create a .yml with same name under directory path `device_specific_store_driver`.
+  - Content of the file should look like:
```
7flying, line 752: "The content" or "The contents"

You can also add a link to the example file that you have added

HOWTO.md
```diff
@@ -734,3 +736,31 @@ Options:
...
+        username: user_device_specific
+        sshkeys:
+        - "testkeyperdevice"
+      files: null
+      commands: null
+      diskencryption_clevis: null
+      additional_serviceinfo: null
+      after_onboarding_reboot: false
+    ```
+  - Follow onboarding procedure and this particular device will get serviceinfo settings as mentioned in above file.
```
7flying, line 764: "Follow the onboarding" .. "will get the serviceinfo" "in the file above"/"in the above file"

HOWTO.md
```diff
@@ -734,3 +736,31 @@ Options:
...
+  Other devices if do not have their specific file under `device_specific_store_driver` they will get onboarded
+  with setting from main file which is `serviceinfo_api_server.yml`.
```
7flying, line 766: You should add this bit of info to the explanation done at the beginning so that the user knows what to expect

integration-tests/tests/common/mod.rs
```diff
@@ -790,10 +797,23 @@ impl<'a> TestServerConfigurator<'a> {
                     "config_dir",
                     &self.test_context.runner_path(&self.server_number),
                 );
-                cfg.insert(
-                    "user",
-                    users::get_current_username().unwrap().to_str().unwrap(),
-                );
+
+                if per_device == "false" {
```
7flying, line 801: it makes more sense to me to get the value from the env var, convert it to boolean and then check this case as a boolean over here instead of comparing it to a string

serviceinfo-api-server/src/main.rs
runcom: I ultimately think this has to be "abstracted" away so that it's easier in the future to switch between serialization/implementation/whatever

serviceinfo-api-server/src/main.rs
runcom: I would just log the error at warning at least and avoid returning the error

serviceinfo-api-server/src/main.rs
sarmahaj: I agree, later on when we consider to override all of the service info (rather than just sshkey &username) the above part is going to be common. So as a part of abstraction I was thinking of moving that to `settings_per_device ` function.

util/src/servers/configuration/serviceinfo_api_server.rs
7flying: Checked the docs, all good, I misunderstood :+1:

serviceinfo-api-server/src/main.rs
sarmahaj: So as discussed , added some level of abstraction to this function, 
  and later when we have Database addition we will consider changes accordingly.

integration-tests/tests/common/mod.rs
sarmahaj: yes above m reading from env variable, added the part to compare with boolean instead of string

integration-tests/tests/common/mod.rs
```diff
@@ -790,10 +800,23 @@ impl<'a> TestServerConfigurator<'a> {
                     "config_dir",
                     &self.test_context.runner_path(&self.server_number),
                 );
-                cfg.insert(
-                    "user",
-                    users::get_current_username().unwrap().to_str().unwrap(),
-                );
+
+                if per_device == false {
```
7flying, line 804: if !per_device

integration-tests/tests/e2e.rs
```diff
@@ -114,6 +124,10 @@ where
     F: Fn(&TestContext) -> Result<(&'static str, String, &'static str)>,
 {
     let ci = env::var("FDO_PRIVILEGED").is_ok();
+    let per_device = env::var("PER_DEVICE_SERVICEINFO")?;
+    L.l(format!("per_device env variable:\n {}", per_device));
+    let per_device = env::var("PER_DEVICE_SERVICEINFO")?;
+    L.l(format!("per_device env variable:\n {}", per_device));
```
7flying, line 130: These two lines are duplicated

sarmahaj: good catch !

integration-tests/tests/e2e.rs
```diff
@@ -114,6 +124,8 @@ where
     F: Fn(&TestContext) -> Result<(&'static str, String, &'static str)>,
 {
     let ci = env::var("FDO_PRIVILEGED").is_ok();
+    let per_device = env::var("PER_DEVICE_SERVICEINFO")?;
+    L.l(format!("per_device env variable:\n {}", per_device));
```
7flying, line 128: To make things clearer, we should fail the test here if `PER_DEVICE_SERVICEINFO` is set to true, since this test has become the non-per-device service info test.
Please rename this function to `test_e2e_impl_standard_serviceinfo` or something like that to better differentiate the functions

also nit: ` L.l(format!("per_device env variable:\n {per_device}"));`

integration-tests/tests/e2e.rs
```diff
@@ -442,3 +430,240 @@ testkey
 
     Ok(())
 }
+
+async fn test_per_device_serviceinfo<F>(
```
7flying, line 434: rename this  one to  `test_e2e_impl_per_device_serviceinfo` or something like that so that we know that is a copy of the original  `test_e2e_impl` with the per-device changes

util/src/servers/mod.rs
```diff
@@ -57,6 +60,44 @@ pub fn settings_for(component: &str) -> Result<config::Config> {
         .context(format!("Loading configuration for {component}"))
 }
 
+pub fn settings_per_device(guid: &str) -> Result<ServiceInfoSettings> {
+    // here we first check if the reuested device has device-specific file stored
```
7flying, line 64: reuested -> requested

integration-tests/tests/common/mod.rs
```diff
@@ -790,10 +800,23 @@ impl<'a> TestServerConfigurator<'a> {
                     "config_dir",
                     &self.test_context.runner_path(&self.server_number),
                 );
-                cfg.insert(
-                    "user",
-                    users::get_current_username().unwrap().to_str().unwrap(),
-                );
+
+                if !per_device {
+                    L.l("per_device_serviceinfo is not set , using default values");
```
7flying, line 805: not set,

7flying: and also something similar below

integration-tests/tests/e2e.rs
```diff
@@ -63,9 +63,12 @@ async fn test_e2e() -> Result<()> {
 
     let mut failed = Vec::new();
 
+    env::set_var("PER_DEVICE_SERVICEINFO", "false");
+
     for (diun_verification_method_name, diun_verification_method_generator) in
         &diun_verification_methods[1..2]
     {
+        //for default serviceinfo
```
7flying, line 71: space after //, as per the rest of the comments in the repo

7flying: in other places too

serviceinfo-api-server/src/main.rs
```diff
@@ -224,12 +227,30 @@ async fn serviceinfo_handler(
         .modules
         .contains(&FedoraIotServiceInfoModule::SSHKey.into())
     {
-        if let Some(initial_user) = &user_data.service_info_configuration.settings.initial_user {
-            reply.reply.initial_user = Some(ServiceInfoApiReplyInitialUser {
-                username: initial_user.username.clone(),
-                ssh_keys: initial_user.sshkeys.clone(),
-            });
-        }
+        // precedence is given to 'per_device' settings over base serviceinfoapi_server.yml config
```
7flying, line 230: serviceinfo_api_server.yml

integration-tests/tests/e2e.rs
sarmahaj: actually I printed this statement during test to check is default env is set to false or not . But yeah the way we set env::set_var("PER_DEVICE_SERVICEINFO", "true"); at the start of test_per_device_serviceinfo() , it will make sense to set it to false at the start of test_e2e_impl_standard_serviceinfo() test. 
And will rename the test too.

HOWTO.md
```diff
@@ -734,3 +736,26 @@ Options:
 
 Please note that in this mode there are some environment variables that are
 still required to be set by the user (`DI_SIGN_KEY_PATH`, `DI_HMAC_KEY_PATH`).
+
+## How to use Features
+
+### How to use `per-device serviceinfo` feature
+
+  Using this feaure the user can choose to apply different serviceinfo settings on different devices.
+  For that the user needs to provide a path to a `per-device serviceinfo` file under the`device_specific_store_driver` field
```
7flying, line 745: space missing after:  ```the`device_```

HOWTO.md
```diff
@@ -734,3 +736,26 @@ Options:
 
 Please note that in this mode there are some environment variables that are
 still required to be set by the user (`DI_SIGN_KEY_PATH`, `DI_HMAC_KEY_PATH`).
+
+## How to use Features
+
+### How to use `per-device serviceinfo` feature
```
7flying, line 742: I think that a "the" is needed in "How to use the `per-device serviceinfo` feature"

HOWTO.md
```diff
@@ -734,3 +736,26 @@ Options:
...
+
+  Using this feaure the user can choose to apply different serviceinfo settings on different devices.
+  For that the user needs to provide a path to a `per-device serviceinfo` file under the`device_specific_store_driver` field
+  present in the `serviceinfo_api_server.yml` file.
+  Other devices if do not have their `per-device serviceinfo` file under `device_specific_store_driver` they will get onboarded
+  with setting from main file which is `serviceinfo_api_server.yml`.
```
7flying, line 748: with settings from the main file,

HOWTO.md
```diff
@@ -734,3 +736,26 @@ Options:
...
+  1. Initialize the device as mentioned in [How to generate an Ownership Voucher and Credential for a Device](#how-to-generate-an-ownership-voucher-ov-and-credential-for-a-device-device-initialization).
+
+  2. Dump the `device-credentials`
+  ```bash
+  fdo-owner-tool dump-device-credential /path/to/device-credentials
+  ```
+
+  3. Note the GUID of the device and create a .yml with same name as the `guid` under directory path `device_specific_store_driver`.
+
+  4. You can refer [per_device_serviceinfo.yml](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/examples/config/device_specific_serviceinfo.yml) as an example.
```
7flying, line 759: refer to

HOWTO.md
```diff
@@ -734,3 +736,26 @@ Options:
...
+
+## How to use Features
+
+### How to use the `per-device serviceinfo` feature
+
+  Using this feaure the user can choose to apply different serviceinfo settings on different devices.
+  For that the user needs to provide a path to a `per-device serviceinfo` file under the `device_specific_store_driver` field
+  present in the `serviceinfo_api_server.yml` file.
+  If other devices do not have their `per-device serviceinfo` file under `device_specific_store_driver` they will get onboarded
+  with setting from the main file, which is `serviceinfo_api_server.yml`.
```
7flying, line 748: sorry an S still missing in settings

7flying (approved):
LGTM :+1:

sarmahaj:
> LGTM 👍

Thanks Irene!

PR #480, opened 2023-04-18
feat(admin-tool): configure X509 certificate's expiration date 

7flying:
Currently with the `generate-key-and-cert` option of the
`fdo-admin-tool` cli we can configure the 'Country' and
'Organization' name of the generated X509 certificate, and we
set by default a 365-day validity period.

This allows to set an specific number of days for the certificate's
validity period with the new `--validity-ends` option.

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -48,28 +48,24 @@ Use `fdo-admin-tool generate-key-and-cert` to generate the required keys for
 `diun`, `manufacturer`, `device-ca` or `owner`. 
 
 ```bash
-USAGE:
-    fdo-admin-tool generate-key-and-cert [OPTIONS] <SUBJECT>
+Usage: fdo-admin-tool generate-key-and-cert [OPTIONS] <SUBJECT>
 
-ARGS:
-    <SUBJECT>    Subject of the key and certificate [possible values: diun, manufacturer,
-                 device-ca, owner]
+Arguments:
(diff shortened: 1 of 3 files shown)


nullr0ute:
What certificates are these for? Are they purely the AIO server side certs? Maybe a more detailed commit message? Also if this is specified via a cli do any of the help descriptions (--help output, man pages etc) need to be updated?

7flying:
@nullr0ute 
> What certificates are these for? Are they purely the AIO server side certs? 

These are the X509 certificates that generate to be used in the `diun, manufacturer, device-ca, owner` uses cases. We also generate them for the `aio` tool. The only fields that we allow to configure right now are the 'Country' and 'Organization', and by default they are issued with a 365-day validity period, but the standard does not say anything about that, so this PR enables a way to specify a number of days the certificate is going to be valid leaving 365 days as the default period.
This is simply a "nice thing" to have since users can issue their own certificates with openssl.

> Maybe a more detailed commit message? 

Handled, yes.

> Also if this is specified via a cli do any of the help descriptions (--help output, man pages etc) need to be updated?

The library that we use automatically handles this based on the code-comments on the Rust structs, I've updated the HOWTO though (missed that).

7flying:
PTA @nullr0ute

nullr0ute (approved):
LGTM

PR #491, opened 2023-05-04
feat(manufacturing-client): use default ipv4 iface route if not provided

### LINKED ISSUE #468: Allow to default to a single available iface when `MACAddress` is used as the `mfg_string_type` (IPv4)
If `mfg_string_type` is set to `MACAddress` we should allow to default to a iface (which isn't loopback) if there is a single one available in the device and the user hasn't told us to get the MAC address from an specific iface.

sarmahaj:
This PR adds to mfg_string_type=MACAddress feat

- in case of mfg_string_type selected as  MACAddress and user has not provided which interface to use then get the default route and use that interface to get MACAddress.
- supports ipv4 network interface
- test is added
- updated HOWTO.md accordingly.
- fixes #468

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -652,18 +652,17 @@ Identification method to the device.
 2. If the Manufacturing server is specifically configured with a
    `mfg_string_type` set to `MACAddress` in its `diun` configuration section it
    will ask the device to identify itself with a MAC Address, in order to do so
-   the device must set `DI_MFG_STRING_TYPE_MAC_IFACE` [optional] to a valid
-   interface.
+   the user has an option to set `DI_MFG_STRING_TYPE_MAC_IFACE` [optional] to
+   a valid interface.
+  Valid interfaces are those which do not result in a `00:00:00:00:00:00` MAC Address.
    
    Please note that this `DI_MFG_STRING_TYPE_MAC_IFACE` environment variable is
    *optional* and will only be read if the server requests a `MACAddress`
-   identification mode, but if the Manufacturing server is configured as
-   described above the Device Initialize protocol will result in an early error
-   if this environment variable is not set.
-   
-   In such case, the value `DI_MFG_STRING_TYPE_MAC_IFACE` must be set to any
-   interface name of the device that does not result in `00:00:00:00:00:00` as
-   the specified MAC Address.
+   identification mode. If the Manufacturing server is configured as
+   described and the environment variable `DI_MFG_STRING_TYPE_MAC_IFACE` is not set
+   then the default active network interface will be used. This is obtained from 
+   kernel's routing table file (`/proc/net/route`).
+  
    
 3. Run the client: `fdo-manufacturing-client`.
 
@@ -675,10 +674,14 @@ identification with the Manufacturing server.
 1. `DI_MFG_STRING_TYPE`: [optional] selects the Device Identification string
    type; by default `serial_number`, other possible value is `mac_address`.
    
-   If `mac_address` is selected as `DI_MFG_STRING_TYPE` then the user *must*
-   specify a valid interface to read the MAC Address from with
-   `DI_MFG_STRING_TYPE_MAC_IFACE`. Valid interfaces are those which do not
-   result in a `00:00:00:00:00:00` MAC Address.
+   If `mac_address` is selected as `DI_MFG_STRING_TYPE` then the user has an option
+   to specify a valid interface to read the MAC Address from with
+   `DI_MFG_STRING_TYPE_MAC_IFACE` [optional] env variable. 
+   Valid interfaces are those which do not result in a `00:00:00:00:00:00` MAC Address.
+   If the user has not specified which network interface to be used then the default 
+   active network interface will be used. This is obtained from kernel's routing table 
+   file (`/proc/net/route`).
+  
    
 2. `DI_KEY_STORAGE_TYPE`: selects the storage type for the Device
    Identification keys, valid values: `filesystem` (`tpm` not yet
--- a/manufacturing-client/src/main.rs
+++ b/manufacturing-client/src/main.rs
@@ -1,6 +1,7 @@
 use anyhow::{bail, Context, Result};
 use clap::{Args, Parser, Subcommand};
 use regex::Regex;
+use std::io::{BufRead, BufReader};
 use std::path::Path;
 use std::{convert::TryFrom, fs};
 use std::{convert::TryInto, env, str::FromStr};
@@ -324,11 +325,26 @@ async fn main() -> Result<()> {
                 url = args.manufacturing_server_url;
 
                 mfg_string_type = args.mfg_string_type;
-                // ensure that we are given the iface if MACAddress is selected
-                if mfg_string_type == MfgStringType::MACAddress && args.iface.is_none() {
-                    bail!("When using MACAddress as the MfgStringType --iface is required");
+                if mfg_string_type == MfgStringType::MACAddress {
+                    // user provided iface
+                    if args.iface.is_some() {
+                        iface = args.iface;
+                    } else {
+                        // If user has not selected any specific iface then default iface will be used
+                        match get_default_network_iface() {
+                            Ok(Some(result)) => {
+                                iface = Some(result);
+                                log::info!("Default network interface found: {iface:#?}");
+                            }
+                            Err(error) => {
+                                bail!("Error retrieving default network interface: {error}");
+                            }
+                            Ok(None) => {
+                                bail!("Error retrieving default network interface, unknown reason");
+                            }
+                        }
+                    }
                 }
-                iface = args.iface;
 
                 keyref = KeyReference::str_key(args.key_ref)
                     .await
@@ -351,15 +367,31 @@ async fn main() -> Result<()> {
                 } else {
                     bail!("No DIUN root key verification methods set");
                 }
-                iface = args.iface;
 
                 log::debug!("Performing DIUN");
                 client = ServiceClient::new(ProtocolVersion::Version1_1, &url);
                 (keyref, mfg_string_type) = perform_diun(&mut client, diun_pub_key_verification)
                     .await
                     .context("Error performing DIUN")?;
-                if mfg_string_type == MfgStringType::MACAddress && iface.is_none() {
-                    bail!("Server has requested mac_address as mfg_string_type and there is no iface set");
+                if mfg_string_type == MfgStringType::MACAddress {
+                    // user provided iface
+                    if args.iface.is_some() {
+                        iface = args.iface;
+                    } else {
+                        // If user has not selected any specific iface then default iface will be used
+                        match get_default_network_iface() {
+                            Ok(Some(result)) => {
+                                iface = Some(result);
+                                log::info!("Default network interface found: {iface:#?}");
+                            }
+                            Err(error) => {
+                                bail!("Error retrieving default network interface: {error}");
+                            }
+                            Ok(None) => {
+                                bail!("Error retrieving default network interface, unknown reason");
+                            }
+                        }
+                    }
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


manufacturing-client/src/main.rs
```diff
@@ -325,10 +327,15 @@ async fn main() -> Result<()> {
 
                 mfg_string_type = args.mfg_string_type;
                 // ensure that we are given the iface if MACAddress is selected
-                if mfg_string_type == MfgStringType::MACAddress && args.iface.is_none() {
-                    bail!("When using MACAddress as the MfgStringType --iface is required");
+                if mfg_string_type == MfgStringType::MACAddress && args.iface.is_some() {
+                    iface = args.iface;
+                } else {
+                    // If user has not selected any specific iface then default iface will be used
+                    iface = get_default_network_iface()?;
```
7flying, line 334: we can also end up in this `else` part if `mfg_string_type == MfgStringType::SeriaNumber`, so  we don't need to get the default iface in that case

manufacturing-client/src/main.rs
```diff
@@ -401,16 +417,19 @@ async fn main() -> Result<()> {
...
-            if env::var("DI_MFG_STRING_TYPE_MAC_IFACE").is_ok() {
-                iface = Some(env::var("DI_MFG_STRING_TYPE_MAC_IFACE").unwrap());
+            if let Ok(iface_var) = env::var("DI_MFG_STRING_TYPE_MAC_IFACE") {
+                iface = Some(iface_var);
             }
             (keyref, mfg_string_type) = perform_diun(&mut client, diun_pub_key_verification)
                 .await
                 .context("Error performing DIUN")?;
             // once diun has been performed we can error ahead of time if the
             // server has requested MACaddress but we haven't set an iface.
```
7flying, line 427: you can delete/rewrite this code comment since it does not apply now after the default iface feature

manufacturing-client/src/main.rs
```diff
@@ -401,16 +417,19 @@ async fn main() -> Result<()> {
             // and it might request MACAddress as the mfg_string_type. What it cannot do
             // is select the iface for the client, so we must set it ahead of time.
             // Thereby, we are just getting this value if provided, hiding errors.
```
7flying, line 419: here if you can delete the `// Thereby .... ` part and write that we are going to try to get a default iface if possible

manufacturing-client/src/main.rs
```diff
@@ -937,3 +956,22 @@ impl KeyReference {
...
+    let file = File::open("/proc/net/route")?;
+    let reader = BufReader::new(file);
+
+    for line in reader.lines().skip(1) {
+        let line = line.unwrap();
+        let fields: Vec<_> = line.split_whitespace().collect();
+
+        // /proc/net/route this is a system file which gives info about kernel's routing table and
+        // the default route has a destination of "00000000". First column gives Iface name.
+        if fields[1] == "00000000" {
```
7flying, line 970: mmm it sounds strange that Clippy did not complain here, what if fields is empty?

sarmahaj: yeah , I tried using clippy::pedantic still it didnt complain!
anyways added a check if fields is empty it will go to next line.

manufacturing-client/src/main.rs
```diff
@@ -937,3 +960,24 @@ impl KeyReference {
...
+    let file = File::open("/proc/net/route")?;
+    let reader = BufReader::new(file);
+
+    for line in reader.lines().skip(1) {
+        let line = line.unwrap();
+        let fields: Vec<_> = line.split_whitespace().collect();
+        if fields.is_empty() {
+            continue;
+        }
+        // /proc/net/route this is a system file which gives info about kernel's routing table and
```
nullr0ute, line 974: What about IPv6?

sarmahaj: Thats a great question!
I will add support to read ipv6 interfaces too.

say-paul:
@sarmahaj  I have accidentally clicked rebase.

manufacturing-client/src/main.rs
sarmahaj: @nullr0ute added support for ipv6 network interfaces, ptal!

manufacturing-client/src/main.rs
```diff
@@ -937,3 +959,52 @@ impl KeyReference {
         }
     }
 }
+
+const IPV6_DEFAULT: &str = "00000000000000000000000000000000"; //DevSkim: ignore DS173237
+const IPV4_DEFAULT: &str = "00000000";
+
+fn get_default_network_iface() -> Result<Option<String>> {
```
puiterwijk, line 966: Why does this return a `Result<_>` if it can't fail?
I'm wondering if the reading of the route tables should return errors if they error out?

puiterwijk (commented):
I wonder if using the default interface during manufacturing is maybe not the best thing to do: this sounds like it could be waiting for "wait, why is this the device ID?" (specifically important for the DI case, since the device ID is used to look up the keys).

Reading the related issue, this is primarily done for the simple case where the device only has a single interface, in which case you could just look through the list of interfaces and pick that, and return "multiple interfaces" if more than one was found.

That would prevent from using the wrong MAC address if a device has multiple interfaces, and uses the wrong one during DI.

sarmahaj:
> I wonder if using the default interface during manufacturing is maybe not the best thing to do: this sounds like it could be waiting for "wait, why is this the device ID?" (specifically important for the DI case, since the device ID is used to look up the keys).
> 
> Reading the related issue, this is primarily done for the simple case where the device only has a single interface, in which case you could just look through the list of interfaces and pick that, and return "multiple interfaces" if more than one was found.
> 
> That would prevent from using the wrong MAC address if a device has multiple interfaces, and uses the wrong one during DI.
@puiterwijk Thanks for reviewing PR !
The idea here was to provide unique device identification , which is done by  mfg_string_type: SerialNumber . This field comes from manufacturer. But if serial number from manufacturer (/sys/devices/virtual/dmi/id/product_serial or /sys/devices/virtual/dmi/ id/chassis_serial) is not available then it used to appear as None , so we came up with this idea => it will be populated as mac addresses of available network interfaces like :
` Device Info: IFACE_eth0=02:42:ac:11:00:03`
And again if user selected mfg_string_type: MACAddress then user needs to provide which interface to be used for the macddress (either as argument or env variable) . If the user has not provided which interface to use then only active network interface(may be in the PR description I should have used "active" if instead of "default" !) is used for this purpose.
Also incase of multiple interfaces , finally we need to select one interface which will be used to get MACAddress as device id. Also as per understanding client wants to use this identification only to physically identify the device (Probably this will be printed on the box or something and then will be used for installing at the correct location. or something !) . We do not use this identification in DI .(I am trying to figure out what you mentioned : "since the device ID is used to look up the keys").)

manufacturing-client/src/main.rs
sarmahaj: it returns either the mac address string or None , if None there was some error determining the interface.
Are you suggesting instead of None, the function should send error ? (for more details!)

7flying: I think that since, for instance, we are not propagating the errors of `std::fs::File::open`, it makes no sense to return `Result`, since there won't be any errors.

If we are going to return `Some` or `None` (like we are doing now), then it would make more sense to return an `Option` instead, or to continue returning `Result`, but propagating the errors that we could encounter.

puiterwijk:
> @puiterwijk Thanks for reviewing PR ! The idea here was to provide unique device identification , which is done by mfg_string_type: SerialNumber . This field comes from manufacturer. But if serial number from manufacturer (/sys/devices/virtual/dmi/id/product_serial or /sys/devices/virtual/dmi/ id/chassis_serial) is not available then it used to appear as None , so we came up with this idea => it will be populated as mac addresses of available network interfaces like : ` Device Info: IFACE_eth0=02:42:ac:11:00:03` And again if user selected mfg_string_type: MACAddress then user needs to provide which interface to be used for the macddress (either as argument or env variable) . If the user has not provided which interface to use then only active network interface(may be in the PR description I should have used "active" if instead of "default" !) is used for this purpose. Also incase of multiple interfaces , finally we need to select one interface which will be used to get MACAddress as device id. Also as per understanding client wants to use this identification only to physically identify the device (Probably this will be printed on the box or something and then will be used for installing at the correct location. or something !) 

Right. But the challenge here is that if a device has two separate interfaces, and another one is used for preparation of DI (and thus expected by the manufacturing server), it'll fail, or at least result in unexpected behaviour (what if the device has a MAC address printed on it for the on-board ethernet port, but during manufacturing, another (USB) interface is plugged in? Or what if the device uses a different Mac, e.g. with (MAC address randomization([https://wiki.archlinux.org/title/NetworkManager#Configuring_MAC_address_randomization])?).
My question is basically: if a device has multiple MAC addresses during manufacturing, should we be okay with just using the one that happened to be in use at this moment, or should we then demand an interface configured?
(Issue #468 says "if there is a single on available", and I guess the question comes down to what is "available". Is it "available" when it's the plugged in one, or when it exists on the device?) 

> We do not use this identification in DI .(I am trying to figure out what you mentioned : "since the device ID is used to look up the keys")

We do, with the DI (non-UN) version of the manufacturing protocol, to look up the key [in the `public_key_store`](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/manufacturing-server/src/handlers/di.rs#L80).
In the DI-UN case, the device public key is provided as part of the -UN part, in the plain DI case, the server is expected to Just Know the device public key in advance.

runcom:
> Right. But the challenge here is that if a device has two separate interfaces

I think this whole thing was a nice to have part which again, is just nice to have. It's difficult and error prone - zezere does something similar too and can just stomp on the same right? maybe we shouldn't push this through or again, leave it as best effort.

manufacturing-client/src/main.rs
sarmahaj: as discussed ,I will create another PR for ipv6, This PR focusses on ipv4 network interface support.

manufacturing-client/src/main.rs
sarmahaj: handled by propagating std error for more visibility on error msg

HOWTO.md
```diff
@@ -647,18 +647,17 @@ Identification method to the device.
...
-   identification mode, but if the Manufacturing server is configured as
-   described above the Device Initialize protocol will result in an early error
-   if this environment variable is not set.
-   
-   In such case, the value `DI_MFG_STRING_TYPE_MAC_IFACE` must be set to any
-   interface name of the device that does not result in `00:00:00:00:00:00` as
-   the specified MAC Address.
+   identification mode. If the Manufacturing server is configured as
+   described and the environment variable `DI_MFG_STRING_TYPE_MAC_IFACE` is not set
+   then the default active network interface will be used. This is obtainbed from
```
7flying, line 658: s/obtainbed/obtained/

manufacturing-client/src/main.rs
```diff
@@ -324,11 +325,23 @@ async fn main() -> Result<()> {
...
+                if mfg_string_type == MfgStringType::MACAddress {
+                    // user provided iface
+                    if args.iface.is_some() {
+                        iface = args.iface;
+                    } else {
+                        // If user has not selected any specific iface then default iface will be used
+                        match get_default_network_iface() {
+                            Ok(result) => {
+                                iface = result;
+                                log::info!("Default network interface found: {:#?}", iface);
```
7flying, line 337: nit: can you change these to `{variable:#?}`, I know that it is a pain but we had clippy raise warnings before

manufacturing-client/src/main.rs
```diff
@@ -937,3 +976,25 @@ impl KeyReference {
...
+        if fields.is_empty() {
+            continue;
+        }
+        if fields[1] == IPV4_DEFAULT && fields[0] != "lo" {
+            let iface = fields[0].to_string();
+            log::info!("Default network interface is ipv4 based {}", iface);
+            return Ok(Some(iface));
+        }
+    }
+    Err(std::io::Error::last_os_error())
```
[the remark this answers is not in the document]
sarmahaj: Right ..so actually the functions returns either result, Err (incase of error opening file) or None , so updated the function and its handling accordingly.

7flying (approved):
LGTM thanks :+1:

sarmahaj:
> LGTM thanks 👍

Thanks for the review :)

7flying:
@Mergifyio refresh

PR #501, opened 2023-05-30
chore: Update libcryptsetup-rs to 0.7

nullr0ute:
Update to the latest libcryptsetup-rs upstream release. Also update the Cargo.lock as part of this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,6 +23,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


nullr0ute:
@7flying can you double check none of the API has changed with this bump in case we need to do more than just bump the rev.

7flying:
> @7flying can you double check none of the API has changed with this bump in case we need to do more than just bump the rev.

@nullr0ute All good, we do not need to change anything on our end.
The commitlint is failing due to an uppercase letter in your commit, other that that we can merge this.

nullr0ute:
> > @7flying can you double check none of the API has changed with this bump in case we need to do more than just bump the rev.
> 
> @nullr0ute All good, we do not need to change anything on our end. The commitlint is failing due to an uppercase letter in your commit, other that that we can merge this.

Should be fixed

7flying (approved):
LGTM

PR #519, opened 2023-06-22
fix: explicitly mark LUKS2 for context load

7flying:
This explicitly marks LUKS2 as the load context to avoid getting `Operation incompatible with device marked for LUKS2 reencryption` errors.

Fixes rhbz#2216111

### DIFF
--- a/client-linuxapp/src/reencrypt/mod.rs
+++ b/client-linuxapp/src/reencrypt/mod.rs
@@ -64,7 +64,7 @@ fn perform_reencrypt(dev_name: &str) -> Result<()> {
         .context("Error opening device")?;
 
     dev.context_handle()
-        .load::<()>(None, None)
+        .load::<libcryptsetup_rs::CryptParamsLuks2Ref>(None, None)
         .context("Error loading device context")?;
 
     let status = dev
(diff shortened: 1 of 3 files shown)


client-linuxapp/src/reencrypt/mod.rs
```diff
@@ -64,7 +64,7 @@ fn perform_reencrypt(dev_name: &str) -> Result<()> {
         .context("Error opening device")?;
 
     dev.context_handle()
-        .load::<()>(None, None)
+        .load::<libcryptsetup_rs::CryptParamsLuks2Ref>(None, None)
```
puiterwijk, line 67:
```suggestion
        .load::<libcryptsetup_rs::CryptParamsLuks2>(None, None)
```

Note: no idea if that'll work, but just a feeling.

7flying: not really, after the bump to 0.8.0 there were some API changes, and that requires specifically specifying which parameters do you want in order to load the header. I changed it to `()` which turned out that was wrong and yielded the bugzilla, the `CryptParamsLuks2Ref` option is what I've tested that works for us

runcom (approved):

Issue #507, opened 2023-06-13
Update e2e tests so that they check the flow with multiple ssh keys per user


7flying:
Right now at the e2e tests we don't test if a user can be created with multiple (more than one) ssh keys associated with it.

The relevant FSIM allows it, but we simply don't test it. So, test it.

PR #522, opened 2023-06-28
chore: update ssh-key format in tests to match real-world usage

7flying:
This change allows us to reproduce [RHBZ#2218231](https://bugzilla.redhat.com/show_bug.cgi?id=2218231) (and to catch more related bugs in the future)

and it also fixes that same BZ.

Our FSIM for creating users with an associated ssh key did
not take into account that ssh keys are strings that may
have spaces, e.g. 'ssh-ed25519 long-string user@example.com',
and used ' ' as the split character.

This sets ';' as the split character in the FSIM avoiding
installing unusable ssh keys.

### DIFF
--- a/client-linuxapp/src/serviceinfo.rs
+++ b/client-linuxapp/src/serviceinfo.rs
@@ -718,7 +718,7 @@ async fn process_serviceinfo_in(si_in: &ServiceInfo, si_out: &mut ServiceInfo) -
             ))?;
             let sshkey_keys_v: Vec<String> = sshkey_keys
                 .unwrap()
-                .split(' ')
+                .split(';')
                 .map(|s| s.to_string())
                 .collect();
             for key in sshkey_keys_v {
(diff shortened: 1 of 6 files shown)


nullr0ute (commented):
One minor nit is the general accepted for non real domains is example.com, I think email.com is a real domain so we may not want to use that.

7flying:
> One minor nit is the general accepted for non real domains is example.com, I think email.com is a real domain so we may not want to use that.

oh, I did not know that, will update this accordingly, thanks :+1:

nullr0ute (approved):
LGTM

7flying:
@nullr0ute sorry about the dismissed review, I ended up fixing the BZ also in this PR

7flying:
Linking #507 for awareness

nullr0ute (approved):
LGTM

PR #525, opened 2023-07-04
Dev dependencies

7flying:
- marks all dependencies in `integration-tests` as dev-dependencies
- changes the Cargo.toml workspace so that by default a `cargo build` does *not* build the integration tests, they will be built if we specify `cargo build --workspace`, meaning that the whole workspace needs to be built
- removes unused dependecies
- removes the `sha-crypt` dependency and uses `openssl` directly. This requires to update the tests since using openssl like this updates the salt of the password every time, whereas with `sha-crypt` we did not do that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,12 +107,12 @@ jobs:
         with:
           toolchain: stable
       - name: Build
-        run: cargo build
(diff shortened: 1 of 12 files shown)


nullr0ute (approved):
LGTM

PR #527, opened 2023-07-05
ci: install jq in "Get Current Job Log URL" step

henrywang:
Package jq is required in "Get Current Job Log URL" step. It should be installed before use it.
Fix error in https://github.com/fedora-iot/fido-device-onboard-rs/actions/runs/5463148796/jobs/9943494222#step:2:9


7flying (approved):
LGTM

PR #528, opened 2023-07-05
test: remove ssh key files and generate them on the fly

henrywang:
This PR also includes a simple version check-ostree.yaml to run FDO testing


puiterwijk (approved):
Cool, thanks!

Issue #531, opened 2023-07-11
Error with `tss_esapi`: Failed to instantiate TCTI


7flying:
From the main branch, running the `manufacturing-client`as root on Fedora 38:
```bash
MANUFACTURING_SERVER_URL=http://192.168.122.180:8080 \
USE_PLAIN_FI=false \
DIUN_PUB_KEY_INSECURE=true \
./target/debug/fdo-manufacturing-client 
 2023-07-11T11:45:37.372Z INFO  fdo_manufacturing_client > No usable device credential located, performing Device Onboarding
 2023-07-11T11:45:37.402Z INFO  fdo_manufacturing_client > Performing DIUN
ERROR:tcti:src/tss2-tcti/tctildr.c:430:Tss2_TctiLdr_Initialize_Ex() Failed to instantiate TCTI 
 2023-07-11T11:45:37.418Z ERROR tss_esapi::tcti_ldr      > Error when creating a TCTI context: response code not recognized
```

But, `device-credentials` are correctly generated.

runcom:
are the credentials "correct" tho?

7flying:
They look OK to me:
```
Active: true
Protocol Version: 101
Device Info: 
Device GUID: ce27e34f-a23e-0593-9489-6d91bf8dcec3
Rendezvous Info:
	- [(DevicePort, [25, 31, 146]), (IPAddress, [68, 192, 168, 122, 180]), (OwnerPort, [25, 31, 146]), (Protocol, [1])]
	- [(DevicePort, [25, 31, 146]), (IPAddress, [80, 254, 128, 0, 0, 0, 0, 0, 0, 151, 226, 23, 22, 106, 168, 136, 186]), (OwnerPort, [25, 31, 146]), (Protocol, [1])]
Public key hash: 576f9815c6e3d8f46d8c9385eb58446a3b4dc258429b1931dc95cddbd8d6b0c81384bc3fb48a98dae965621019d4df58 (Sha384)
HMAC and signing key:
	HMAC key: <secret>
	Signing key: <secret>
```

7flying:
If we isolate the lines of code that cause the error:
```rust
 let tcti_conf = tss_esapi::tcti_ldr::TctiNameConf::Tabrmd(Default::default());
 let _tss_context =
        tss_esapi::Context::new(tcti_conf).context("Error initializing the TPM context")?;
```

we get: 

```
[...]

PR #532, opened 2023-07-12
[WIP] chore: test device credential generation with TPM

7flying:
Added a test to check if we can reproduce #531 upstream.

### DIFF
--- a/integration-tests/tests/common/mod.rs
+++ b/integration-tests/tests/common/mod.rs
@@ -741,6 +741,13 @@ impl TestClientResult {
     pub fn raw_stdout(&self) -> &[u8] {
         &self.raw_stdout
     }
+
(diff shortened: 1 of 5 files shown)


7flying:
We do get the same [error](https://github.com/fedora-iot/fido-device-onboard-rs/actions/runs/5534073627/jobs/10098530554?pr=532#step:9:449) :

```
Common: ERROR:tcti:src/tss2-tcti/tctildr.c:430:Tss2_TctiLdr_Initialize_Ex() Failed to instantiate TCTI 
Common:  2023-07-12T16:33:47.668Z ERROR tss_esapi::tcti_ldr                   > Error when creating a TCTI context: response code not recognized
```

plus some extra stuff

PR #533, opened 2023-07-14
[WIP] chore: Update e2e tests to check flow with multiple ssh keys per user

### LINKED ISSUE #507: Update e2e tests so that they check the flow with multiple ssh keys per user

rdotjain:
This PR attempts to test if a user can be created with multiple ssh keys associated with it.

- Updates tests to check user creation with multiple ssh keys.
- Fixed clippy `using .borrow() on double reference` warnings.

closes #507

### DIFF
--- a/integration-tests/tests/common/mod.rs
+++ b/integration-tests/tests/common/mod.rs
@@ -807,7 +807,8 @@ impl<'a> TestServerConfigurator<'a> {
                         "user",
                         users::get_current_username().unwrap().to_str().unwrap(),
                     );
-                    cfg.insert("sshkey", "ssh-ed25519 sshkey_default user@example.com");
+                    cfg.insert("sshkey1", "ssh-ed25519 sshkey_default user@example.com");
+                    cfg.insert("sshkey2", "ssh-ed25519 sshkey_default user@example2.com");
                     cfg.insert("password", "testpassword");
                 } else {
                     L.l("per_device_serviceinfo is set, using device specific values");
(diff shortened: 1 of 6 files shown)


integration-tests/tests/common/mod.rs
```diff
@@ -807,15 +807,15 @@ impl<'a> TestServerConfigurator<'a> {
...
+                    cfg.insert("sshkey", "ssh-ed25519 sshkey_default user@example.com;ssh-ed25519 sshkey_default user@example2.com");
                     cfg.insert("password", "testpassword");
                 } else {
                     L.l("per_device_serviceinfo is set, using device specific values");
                     cfg.insert(
                         "user",
                         users::get_current_username().unwrap().to_str().unwrap(),
                     );
-                    cfg.insert("sshkey", "ssh-ed25519 sshkey_per_device user@example.com");
+                    cfg.insert("sshkey", "ssh-ed25519 sshkey_per_device user@example.com;ssh-ed25519 sshkey_default user@example2.com");
```
7flying, line 818: this inserts a string in the configuration file that does not match what a user would do:
```yaml
service_info:
  initial_user:
    username: testuser
    password: testpassword
    sshkeys:
    - ssh-ed25519 sshkey_per_device user@example.com;ssh-ed25519 sshkey_default user@example2.com
```

we are defining a [list](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/util/src/servers/configuration/serviceinfo_api_server.rs#L81C1) in there, so something like the following would be desirable:
```yaml
service_info:
  initial_user:
    username: testuser
    password: testpassword
    sshkeys:
    - ssh-ed25519 sshkey_per_device user@example.com
    - ssh-ed25519 sshkey_default user@example2.com
```

7flying (commented):
Looks good; I would like a description added to the `chore: test flow with multiple ssh keys per user`commit, so that we can all understand what `flow` means here. In the clippy fix simply replace `error` with `warnings`.

7flying (approved):
LGTM

PR #536, opened 2023-07-21
Change TCTI type from `Tabrmd` to `Device` (Kernel RM)

7flying:
We were using `Tabrmd` as the way to communicate with the TPM, however we should be using the Kernel resource monitor instead. This PR changes that and we default to using `/dev/tpmrm0` if no other named config is passed via
environment variables (TPM2TOOLS_TCTI, TCTI, TEST_TCTI). 

This change is documented in the HOWTO.
This PR also adds a test checking that device credentials can be generated with a TPM.

We also disable using TPMs by default when using the `aio` tool as it now requires using the kernel resource monitor which is not always available.


integration-tests/tests/di_diun.rs
```diff
@@ -183,3 +184,60 @@ async fn get_valid_iface_and_mac() -> Result<(String, String)> {
...
+async fn test_device_credentials_with_tpm() -> Result<()> {
+    let ci = env::var("FDO_PRIVILEGED").is_ok();
+    if !ci {
+        L.l("Skipped test_device_credentials with TPM\nTo run this tet set env variable FDO_PRIVILEGED and run test as superuser");
+        return Ok(());
+    }
+    if std::fs::File::open("/dev/tpm0").is_ok() {
+        env::set_var("TEST_TCTI", "device:/dev/tpm0");
+    } else {
+        L.l("Skipped test_device_credentials with TPM\nTo run this test we need /dev/tpm0");
```
nullr0ute, line 198: Should we be testing with tpmrmX or is this test explicitly for the abrmd and we expand/add another test for the kernel RM?

HOWTO.md
```diff
@@ -289,7 +289,12 @@ Where:
       [here](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/util/src/servers/configuration/manufacturing_server.rs#L71). 
     - `allowed_key_storage_types`: list of allowed storage types. Possible
       values: `FileSystem`, `Tpm` (up-to-date list of options
-      [here](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/util/src/servers/configuration/manufacturing_server.rs#L86))
+      [here](https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/util/src/servers/configuration/manufacturing_server.rs#L86)).
+      
+      In order to use the `Tpm` option you must have the kernel TPM 2 resource
+      manager (`/dev/tpmrm0`) available, or you must set your TPM 2
```
nullr0ute, line 295: Minor nit you use monitor in the commit text and manager here, update the commit to s/monitor/manager as manager is the correct term (there may have been another use of monitor somewhere else).

integration-tests/tests/di_diun.rs
7flying: I haven't added the tpmrmX test since it doesn't work when we are using containers (as we do in Github's CI), it changes the permissions to the following:
```
crw-rw----.  1 nobody nobody  10,   224 Jul 21 08:57 tpm0
crw-rw----.  1 nobody nobody 253, 65536 Jul 21 08:57 tpmrm
```
which gives errors. 

It does work if you are using a VM and add a TPM device, but once you are using containers it breaks.
Hence, just doing a test with the software TPM to at least have something.

HOWTO.md
7flying: thanks, will do :+1:

7flying:
Re-tested, all good on my end, marked as ready to review again.

nullr0ute (approved):
LGTM

PR #538, opened 2023-07-27
fix: drop unused sha-crypt dep

nullr0ute:
The use of sha-crypt was dropped with commit 8d1d1b2 but one of the Cargo.toml updates was missed so drop it there and update Cargo.lock to match.

Fixes: 8d1d1b2 ("chore: replace sha-crypt with openssl process calls")


7flying (approved):
Good catch :+1:

nullr0ute:
> Good catch +1

There's always one :rofl:

PR #539, opened 2023-07-27
feat: replace shadow and passwd crates with our built-in functions

7flying:
This removes the dependencies on shadow and passwd from the project since we can accomplish the same functionality reading the /etc/shadow and /etc/passwd files.


nullr0ute (approved):
LGTM

PR #540, opened 2023-08-02
fix: use J option (xz) instead of j (bzip2) for compression

7flying:
We are setting a .tar.xz extension so we should build a .tar.xz instead of a bzip2. This sets such option in `make-vendored-tarfile.sh`


nullr0ute (approved):
Good catch on my fat finger, not detected as "tar xf" will auto detect the compression :)

PR #541, opened 2023-08-10
fix(spec): add both clients as fdo-admin-cli's Requires

7flying:
We removed the clients from the Requires list so the `fdo-aio.service` can't start. This fixes that.

Fixes rhbz#2230537

### DIFF
--- a/fido-device-onboard.spec
+++ b/fido-device-onboard.spec
@@ -238,6 +238,8 @@ Requires: fdo-manufacturing-server = %{version}-%{release}
 Requires: fdo-rendezvous-server = %{version}-%{release}
 Requires: fdo-owner-onboarding-server = %{version}-%{release}
 Requires: fdo-owner-cli = %{version}-%{release}
+Requires: fdo-client = %{version}-%{release}
+Requires: fdo-init = %{version}-%{release}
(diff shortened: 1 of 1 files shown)


nullr0ute:
Why does the FDO AIO require the clients, it's server side there shouldn't be a dep here?

7flying:
> Why does the FDO AIO require the clients, it's server side there shouldn't be a dep here?

@nullr0ute because the aio tool also has the ability to generate device credentials and start the onboarding, it has the client-linuxapp and the manufacturing-client as dependencies in the code. (I haven't ever used these options)

```
$ fdo-admin-tool aio device --help
Usage: fdo-admin-tool aio device --device-credential-location <DEVICE_CREDENTIAL_LOCATION> <COMMAND>

Commands:
  manufacture  
  analyze      
  run          
  help         Print this message or the help of the given subcommand(s)

Options:
      --device-credential-location <DEVICE_CREDENTIAL_LOCATION>  
  -h, --help                                                     Print help
  -V, --version                                                  Print version
```

runcom:
lgtm - @nullr0ute please take another look

nullr0ute (approved):

PR #547, opened 2023-08-30
chore: fixup paths for selinux

runcom:
Using the service-info-api server with selinux requires us to put anything we want to send to the device under /var/lib/fdo as that directory, and its files, will now get the correct selinux label. The previous approach opens up for security issues by leaving the process basically accessing the whole host.

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -446,10 +446,10 @@ service_info:
     sshkeys:
     - "testkey"
   files:
-  - path: /device/etc/hosts
+  - path: /var/lib/fdo/service-info-api/files/hosts
     permissions: 644
     source_path: /server/local/etc/hosts
-  - path: /device/etc/resolv.conf
+  - path: /var/lib/fdo/service-info-api/files/resolv.conf
     source_path: /server/local/etc/resolv.conf
   commands:
   - command: ls
@@ -493,7 +493,7 @@ Where:
   - `files`: [OPTIONAL] transfers files to a device.
     - `path`: destination path.
     - `permissions`: permissions to set on the file.
-    - `source_path`: source file path.
+    - `source_path`: source file path, must be a file under `/var/lib/fdo/`.
   - `commands`: [OPTIONAL] executes the given list of commands on the device.
       - `command`: command to execute.
       - `args`: list of arguments for the command.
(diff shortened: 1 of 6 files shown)


7flying:
Can you also update the example on https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/HOWTO.md?plain=1#L448-L454 and add a comment on https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/HOWTO.md?plain=1#L493-L494 ?

HOWTO.md
```diff
@@ -446,10 +446,10 @@ service_info:
     sshkeys:
     - "testkey"
   files:
-  - path: /device/etc/hosts
+  - path: /var/lib/fdo/etc/hosts
     permissions: 644
     source_path: /server/local/etc/hosts
-  - path: /device/etc/resolv.conf
+  - path: /var/lib/fdo/etc/resolv.conf
```
nullr0ute, line 452: I think we would want anything the service-info API sends to be in a sub directory under /var/lib/fdo and not in the root directory. Ultimately we're going to end up with DBs and potentially a raft of other things and we likely don't want that pouted with a bunch of random files.

nullr0ute: So likely something like /var/lib/fdo/service-info-api/files/

runcom: there's nothing enforcing that but I can definitely update the examples

nullr0ute: right, I think we should be setting a decent example here, and possibly in the next update actually doing more setup around directories, but that's a different RFE.

runcom:
fixed, should be good now

HOWTO.md
```diff
@@ -446,10 +446,10 @@ service_info:
     sshkeys:
     - "testkey"
   files:
-  - path: /device/etc/hosts
+  - path: /var/lib/fdo/service-info-api/files/hosts
     permissions: 644
     source_path: /server/local/etc/hosts
-  - path: /device/etc/resolv.conf
+  - path: / /var/lib/fdo/service-info-api/files/resolv.conf
```
7flying, line 452: there are two `/` here

7flying:
You can cherry-pick the commit from #548 (or we can merge and rebase) to make the clippy test pass

HOWTO.md
runcom: damn lol

runcom:
> You can cherry-pick the commit from #548 (or we can merge and rebase) to make the clippy test pass

done

7flying (approved):
LGTM :+1:

PR #548, opened 2023-08-31
fix(data-formats): fixes redundant pattern matching

7flying:
There has been a new clippy update, which has triggered new warnings, this fixes those warnings to unblock PRs.


runcom:
cherry-picked in https://github.com/fedora-iot/fido-device-onboard-rs/pull/547

Issue #477, opened 2023-04-17
`fdo-manufacturing-client plain-di` does work


nak3:
As described in https://fedora-iot.github.io/fido-device-onboard-rs/specs/diun.html,
`manufacturing-server` needs to get public key prior to the start of the protocol.

Currently `manufacturing-server` tries to load the public key here:

https://github.com/fedora-iot/fido-device-onboard-rs/blob/69fec227de04d98c51a5b5d8eb218bdc264f0a3f/manufacturing-server/src/handlers/di.rs#L77-L84

but there are several problems for it.

1. The file path is `{public_key_store}/{msg_info}`, which `msg_info` is not a pub key file name. It should allow uses to set the path like `public_key_path` via configfile.
2. It calls `store.load_data()` to load the public key but `load_data()` deserialize the public key. It is not practical that users need to set the public key with the format change.

Also, there is no integration test to reveal `fdo-manufacturing-client plain-di` does not work.

7flying:
1. We already configure it with a configuration file (https://github.com/fedora-iot/fido-device-onboard-rs/blob/main/examples/config/manufacturing-server.yml#L10); the `mfg_info` there specifies the file inside that directory, which matches a device's serial number or MAC address, and it's what we use to uniquely identify a device.
2. 
3. I agree, it expects cbor and makes the rest fail because we are supplying plain data

PR #553, opened 2023-09-07
Deserialize plain DI certs as raw DER

puiterwijk:
Because of the deserialize implementation that's automatically generated, at this moment the expected value for the public_key_store in the manufacturing server is a CBOR array of the DER certificate. This commit adds a new type PlainBytes which (de)serializes transparently, and makes the manufacturing server use it for the public key store.

NOTE: this means that with this patch, the store format on disk changes. This store is a ReadOnly (the server will never write to it), but if anyone would've put a CBOR file in place, that will now fail to open. Raw DER was always the intention (and documented) format, but it still is a risk.

Fixes: #477

### DIFF
--- a/data-formats/src/cborparser.rs
+++ b/data-formats/src/cborparser.rs
@@ -5,7 +5,7 @@ use thiserror::Error;
 
 use crate::{
     constants::HashType,
-    serializable::{DeserializableMany, MaybeSerializable},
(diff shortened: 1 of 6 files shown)


7flying:
tested, works as expected, @puiterwijk you've got a commit lint warning, can you rename your commit to something like "fix: deserialize plain DI certs as raw DER", thanks!

PR #556, opened 2023-09-29
feat: add DB initialization module

7flying:
This provides an interface defining the methods that we expect to be used for OV handling operations.

Added an implementation for sqlite and postgress. I have been unable to make the implementation more generic (aka just one definition of each of the functions of the lib) with a trait with two generic types instead of one, because the `Connection` trait (which holds `SqliteConnection` and `PgConnection`) cannot be made into an object. But I'm happy to learn how to refactor it if someone has ideas.

Added db creation files and tests, ~~the PR is marked as draft since I still need to wire the tests to our upstream CI, making just the sqlite dependencies required, we don't want to install a postgresql server.~~

**Update:** This also adds the capability to select the DB storage method in the relevant server configuration files, the store interface implementation, and two new functions in the fdo-owner-cli in order to export OVs from the manufacturing server and import OVs into the owner-onboarding server.


djach7 (commented):
Lgtm, reviewed for English conventions and with limited Rust knowledge.

migrations_postgres/2023-09-07-131701_create_db/up.sql
```diff
@@ -0,0 +1,22 @@
...
+    ov_guid varchar(36) PRIMARY KEY references ownership_voucher(guid) ON DELETE CASCADE,
+    to2_performed boolean,
+    to0_accept_owner_wait_seconds bigint
+);
+
+CREATE TABLE rendezvous_vouchers (
+    ov_guid varchar(36) PRIMARY KEY references ownership_voucher(guid) ON DELETE CASCADE,
+    ttl bigint
+);
+
```
runcom, line 22: what if instead of 3 more tables, we have attributes only in the ownership_voucher table? the fact that to2 has been performed is a field of the OV itlself, same for the owner_wait_seconds and the ttl once it's sent to the rendezvous - I have not yet thought how this would look like but just something to think about (I came up with this questions because it seems a repetition when we could instead play with attributes on the OV itself as that's what we really deal with)

7flying: I went with that option first, but if we separate things by table we are going to be able to give separate permissions and allow each server to access just what it needs, there are fields that shouldn't be touched by other servers.
We can handle those permissions if we have separate tables, but (to the best of my knowledge) we cannot do that easily if we have a single table.

Also, those fields are not OV fields "inside the OV", but extended attributes that we add to the OV file. Those extended attributes only appear on the OV when it is stored in a specific server, aka ttl is just on the rendezvous server and so on.

runcom: and at this point, if the above is true, then we'd need to have a schema for each server, at which point, it might as well just be called ownership_voucher for all the servers - what bothers me is that we are defining just one schema with tables named differently for things that won't possibly even run on the same db.
For instance, I want to act only as rendezvous server, I don't need the "extra" tables and I only want to write OV+ttl - I might as well just have a table named ownership_voucher anyway but we're bundling everything together like this :thinking:

7flying: yep. I was making the assumption of a single database server running somewhere were all the servers could connect to. But that is not something that works if we want everything to be independent. So, we need three different database initialization scripts since there are going to be as many databases as server types (maybe owner+service info api are a single entity). And there, yes a single table with the fields that each server needs.

7flying: thanks for the comments, I was making a wrong assumption :pray: 
Will update accordingly

runcom: cool cool - I actually raised this up for discussion - my assumption is that servers can run completely indipendenly from each other (surely for manufacturer and the rest) - we can discuss it further if we think that's not the case

migrations_sqlite/2023-09-07-131701_create_db/up.sql
```diff
@@ -0,0 +1,24 @@
+-- Your SQL goes here
+
+CREATE TABLE ownership_voucher (
+    guid varchar(36) NOT NULL PRIMARY KEY,
+    contents blob NOT NULL
+);
+
+CREATE TABLE manufacturer_vouchers (
+    ov_guid varchar(36) NOT NULL PRIMARY KEY, 
+    FOREIGN KEY (ov_guid) REFERENCES ownership_voucher(guid) ON DELETE CASCADE
```
say-paul, line 10: should the same field be both primary and foreign key?

db/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Gets an OwnershipVoucher from the DB given its Guid
+    fn get_ov(guid: &str, conn: &mut T) -> Result<OV>;
+
+    /// Deletes an ownership voucher from the DB given its Guid
+    fn delete_ov(guid: &str, conn: &mut T) -> Result<()>;
+
+    /// Inserts an OV reference in the rendezvous server table
+    fn insert_ov_ref_rv(guid: &str, ttl: Option<i64>, conn: &mut T) -> Result<()>;
+
+    /// Updates the ttl of an OV in the rendezvous server table
```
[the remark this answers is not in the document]
runcom: time to live

migrations_sqlite/2023-09-07-131701_create_db/up.sql
7flying: yes

db/src/lib.rs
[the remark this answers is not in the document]
7flying: time to live

7flying:
Moving this PR back to draft since we need separate sql initialization files.

7flying:
Tests blocked due until  https://github.com/puiterwijk/rust-openssl-kdf/pull/16 goes in

7flying:
Force pushed to get the openssl-kdf update in.

7flying:
Onboarding with DB sqlite implementation done, I need to take care of a couple of things: warnings on CI, add more tests, rewrite some commit messages (maybe merge some of the last commits). Then add the postgreslq part.

7flying:
I'm moving some commits around and merging some others to make the PR more readable.

7flying:
Added [`fix(data-formats): error when trying to parse an empty slice`](https://github.com/fdo-rs/fido-device-onboard-rs/pull/556/commits/fda80e108c8af10a9d3893d0cd33d67a858d391a) and moved it up before the fdo-owner-tool additions since the empty slice check is related to the new OV import cli method.

7flying:
@nullr0ute I have a packaging question: we will need to add all the new `migrations_*` folders to the different RPMs (those folders hold, for each database and server type, the sql files that initialize the database), where would be place those folders?

7flying:
Rebased on main, added docs.

7flying:
Fixed some typos

henrywang:
@7flying Test for postgres added. Could you please give it a review? Thanks.

7flying:
> @7flying Test for postgres added and passed. Could you please give it a review? Thanks.

LGTM, we are testing what we need, thanks :+1:

7flying:
Rebased on main, changed .sql files' install dir to `%{_localstatedir}/lib/fdo/migrations`

7flying:
> Rebased on main, changed .sql files' install dir to `%{_localstatedir}/lib/fdo/migrations`

reverted that to `%{_docdir}/fdo`

7flying:
this force-push reverts some of David's changes since he unknowingly deleted my new commit with his force-push, he will add his new changes later

7flying:
Force pushed to update the docs commit, all ready.

owner-tool/src/main.rs
```diff
@@ -99,6 +109,34 @@ struct ExtendOwnershipVoucherArguments {
     new_owner_cert: String,
 }
 
+#[derive(Args)]
+struct ExportManufacturerVouchersArguments {
+    /// Type of the Manufacturer DB holding the OVs
+    db_type: DBType,
+    /// DB connection URL, or path to the DB file
```
djach7, line 116: English nit: remove the comma here, should be _URL or path_

owner-tool/src/main.rs
```diff
@@ -581,3 +621,169 @@ fn extend_voucher(args: &ExtendOwnershipVoucherArguments) -> Result<(), Error> {
...
+        }
+    }
+    Ok(())
+}
+
+fn import_ownership_vouchers(args: &ImportOwnershipVouchersArguments) -> Result<()> {
+    let source_path = Path::new(&args.source_path);
+    let mut error_buff: Vec<String> = vec![];
+    if source_path.is_dir() {
+        // Import all the ovs in a directory, we will read them one by one and
```
djach7, line 687: English nit: capitalize _OV_ to bring in line with other comments

djach7 (approved):
Lgtm, reviewed for English conventions and tested locally in dev container.

7flying:
The force-push addresses @djach7 comments

djach7 (dismissed):
Lgtm, reviewed for english conventions with some Rust knowledge. Ran cargo tests locally and in dev container too.

mmartinv (commented):
There are some things that look a bit strange to me like putting Manufacturer, Owner and Rendevouz traits and implementations in what should be a place for common things only. I would move all that code to the corresponding server dirs and leave the common code, if any, in the `db` directory.

7flying:
> There are some things that look a bit strange to me like putting Manufacturer, Owner and Rendevouz traits and implementations in what should be a place for common things only. I would move all that code to the corresponding server dirs and leave the common code, if any, in the `db` directory.

Due to the way in which we control how the storage methods are selected in the configuration files, the common code needs to be placed in the `store` module, not in the `db` module, so that's why the common code is already there.
The rest of the code, which are the sqlite and postgres implementations, is in a different module so that we can control via the module features which things we can turn on or off.

HOWTO.md
```diff
@@ -212,6 +215,61 @@ Use `fdo-owner-tool dump-ownership-voucher`:
...
+  [GUID]     GUID of the voucher to be exported, if no GUID is given all the OVs will be exported
+```
+
+For example:
+
+```bash
+fdo-owner-tool export-manufacturer-vouchers postgres \
+postgresql://test:test@localhost/test_manufacturer \
+/path/to/manufacturer-exports/
+```
```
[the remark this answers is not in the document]
7flying: I added the export-import tools to the owner tool since the owner is the one that owns the OVs, and similar methods to handle OVs and device credentials are in the owner tool CLI, not in the admin tool.
The admin tool is only used to generate certs and run the aio tool, it does not have any OV handling capabilities.

HOWTO.md
```diff
@@ -274,7 +332,33 @@ Where:
...
+        Directory:
+          path: /home/fedora/ownership_vouchers
+      ```
+    - `Sqlite`: will use a Sqlite database to store the ownership vouchers.
+      When using this option you must set `Manufacturer` as the DB type as
+      shown below:
+      ```
+      ownership_voucher_store_driver:
+        Sqlite:
+          Manufacturer
```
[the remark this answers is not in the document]
7flying: So, for the first part, regarding that you need to specify the database type, I do agree, but it is a bit complicated, but I'll look into it again.

For the postgres-specific part, I went with no configuration parameters since the amount of parameters varies based on how you have configured your server (e.g. password is not always needed, sometimes you need a timeout etc), and in order to make the postgres and sqlite configurations alike, I went with the environment variable option. We also use environment variables to configure some clients, so I'm OK with it.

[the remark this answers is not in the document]
mmartinv: > So, for the first part, regarding that you need to specify the database type, I do agree, but it is a bit complicated, but I'll look into it again.
> 
> For the postgres-specific part, I went with no configuration parameters since the amount of parameters varies based on how you have configured your server (e.g. password is not always needed, sometimes you need a timeout etc), and in order to make the postgres and sqlite configurations alike, I went with the environment variable option. We also use environment variables to configure some clients, so I'm OK with it.

Another possibility, which might be simpler, would be something like:
```yaml
ownership_voucher_store_driver:
  Database:
    URL: "file:///path/to/the/sqlite/db?option1=value1&option2=value2"
```
or
```yaml
ownership_voucher_store_driver:
  Database:
    URL: "postgresql://username:password@host:5432/database_name?option1=value1&option2=value2"
```

.devcontainer/devcontainer.json
```diff
@@ -28,5 +39,6 @@
 		"cargo",
 		"build"
 	],
+	"postCreateCommand": "cargo install --force diesel_cli --no-default-features --features sqlite && diesel migration run --migration-dir ./migrations/migrations_manufacturing_server_sqlite --database-url ./ci-manufacturer-db.sqlite && diesel migration run --migration-dir ./migrations/migrations_owner_onboarding_server_sqlite --database-url ./ci-owner-db.sqlite && diesel migration run --migration-dir ./migrations/migrations_rendezvous_server_sqlite --database-url ./ci-rendezvous-db.sqlite",
```
runcom, line 42: so everytime we rebuild the container we run the migrations to populate the db right? (just for my own knowledge)

7flying: CC'ing @djach7

test/fdo-postgres.sh
```diff
@@ -0,0 +1,196 @@
...
+  "localhost/manufacturing-server:latest"
+
+greenprint "🔧 Starting fdo owner onboarding server"
+sudo podman run -d \
+  --ip "$FDO_OWNER_ONBOARDING_ADDRESS" \
+  --name owner-onboarding-server \
+  --network edge \
+  -v "$PWD"/test/fdo/:/etc/fdo/:z \
+  -p 8081:8081 \
+  -e POSTGRES_OWNER_DATABASE_URL="postgresql://${POSTGRES_USERNAME}:${POSTGRES_PASSWORD}@${POSTGRES_IP}/${POSTGRES_DB}" \
```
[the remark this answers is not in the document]
7flying: CC-ing @henrywang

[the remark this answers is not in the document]
7flying: But I don't get your question @say-paul , what do you mean by all-server?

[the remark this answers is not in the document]
say-paul: we can specify the Database url in server configs directly instead of passing it as env variable.

say-paul (approved):
LGTM.

runcom (approved):

PR #577, opened 2023-11-17
fix: integration-tests: remove users crate

### LINKED ISSUE #566: Replace the `users` crate with `sysinfo`
The `users` crate that we currently use in our integration tests is unmaintained and security patches are not being applied. Ditch that crate in favour of [sysinfo](https://crates.io/crates/sysinfo).

There is a direct fork of `users` called [uzers](https://crates.io/crates/uzers), but that crate is not packaged in Fedora, and `sysinfo` is, so the change seems easier this way.

djach7:
The users crate has been unmaintained for some time. This fix replaces its functionality with terminal commands resulting in the same output, and removes the users crate from the dependencies.

Fixes #566

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1511,7 +1511,6 @@ dependencies = [
  "tempfile",
  "tera",
  "tokio",
- "users",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


7flying (commented):
Fantastic, this removes a dependency and we still have the same functionality that we needed, will approve once you mark the PR as ready.

7flying (approved):
LGTM

PR #578, opened 2023-11-21
add HTTPS support to owner onboarding server

rdotjain:
- made upload/delete endpoints for ov management api function
- added tests for ov management api
- added HTTPS support to owner server

To-Do:
- add CI Tests

### DIFF
--- a/owner-onboarding-server/src/main.rs
+++ b/owner-onboarding-server/src/main.rs
@@ -1,6 +1,5 @@
 use std::convert::{TryFrom, TryInto};
 use std::fs;
-use std::net::SocketAddr;
 use std::sync::Arc;
 
 use anyhow::{bail, Context, Result};
@@ -38,12 +37,13 @@ use std::convert::Infallible;
 use tls_listener::TlsListener;
 
 pub mod tls_config;
+use tls_config::tls_acceptor;
 
 mod handlers;
 mod ov_management;
 use crate::ov_management::ov_filter;
 
-pub(crate) struct OwnerServiceUD {
+pub struct OwnerServiceUD {
     // Trusted keys
     #[allow(dead_code)]
     trusted_device_keys: X5Bag,
@@ -58,6 +58,8 @@ pub(crate) struct OwnerServiceUD {
         >,
     >,
     session_store: Arc<fdo_http_wrapper::server::SessionStore>,
+    owner_server_https_cert: AbsolutePathBuf,
+    owner_server_https_key: AbsolutePathBuf,
 
     // Our keys
     owner_key: PKey<Private>,
@@ -285,8 +287,8 @@ async fn main() -> Result<()> {
         .context("Error parsing configuration")?;
 
     // Bind information
-    let bind_addr = settings.bind.clone();
-    let addr = SocketAddr::from(([0, 0, 0, 0], 8081));
+    let bind_http_addr = settings.bind_http.clone();
+    let bind_https_addr = settings.bind_https.clone();
 
     // Trusted keys
     let trusted_device_keys = {
@@ -352,6 +354,8 @@ async fn main() -> Result<()> {
         // Stores
         ownership_voucher_store,
         session_store: session_store.clone(),
+        owner_server_https_cert: settings.owner_server_https_cert,
+        owner_server_https_key: settings.owner_server_https_key,
 
         // Trusted keys
         trusted_device_keys,
@@ -424,23 +428,23 @@ async fn main() -> Result<()> {
     let routes = warp::post()
         .and(
             hello
-                .or(handler_ping)
-                .or(handler_report_to_rendezvous)
+                .or(handler_ping.clone())
+                .or(handler_report_to_rendezvous.clone())
                 // TO2
-                .or(handler_to2_hello_device)
-                .or(handler_to2_get_ov_next_entry)
-                .or(handler_to2_prove_device)
-                .or(handler_to2_device_service_info_ready)
-                .or(handler_to2_device_service_info)
-                .or(handler_to2_done)
+                .or(handler_to2_hello_device.clone())
+                .or(handler_to2_get_ov_next_entry.clone())
+                .or(handler_to2_prove_device.clone())
+                .or(handler_to2_device_service_info_ready.clone())
+                .or(handler_to2_device_service_info.clone())
+                .or(handler_to2_done.clone())
                 .or(ov_filter(user_data.clone())),
         )
         .recover(fdo_http_wrapper::server::handle_rejection)
-        .with(warp::log("owner-onboarding-service"));
+        .with(warp::log(
+            "owner-onboarding-service to handle http and https",
+        ));
 
-    log::info!("Listening on {}", addr);
-
-    let service = warp::service(routes);
+    let service = warp::service(routes.clone());
 
     let make_svc = hyper::service::make_service_fn(move |_| {
         let svc = service.clone();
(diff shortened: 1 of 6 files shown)


owner-onboarding-server/src/main.rs
```diff
@@ -423,32 +432,48 @@ async fn main() -> Result<()> {
...
-        })
-        .1;
-    let server = tokio::spawn(server);
-
-    tokio::select!(
-    _ = server => {
-        log::info!("Server terminated");
-    },
-    _ = maintenance_runner => {
-        log::info!("Maintenance runner terminated");
```
7flying, line 450: This handles the core functionality of the server according to the FDO standard, we cannot delete it.

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+async fn add_ov(
+    count: HeaderValue,
+    content_type: HeaderValue,
+    body: impl Buf,
+    user_data: Arc<OwnerServiceUD>,
+) -> core::result::Result<Box<dyn warp::reply::Reply>, warp::Rejection> {
+    // check for ov type
+    let content_type = content_type.to_str().unwrap();
```
7flying, line 54: generally, we cannot bail on a server, we should return an error and continue listening to requests

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+    let mut reader = body.reader();
+    //max limit of ov size is 1024 * 32 bytes
+    let mut dst = [0; 32768];
+    let _ = reader.read(&mut dst).unwrap();
```
7flying, line 58: same as above

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    let ovs = match content_type {
+        "application/x-pem-file" => OwnershipVoucher::many_from_pem(&dst[..]),
+        "application/cbor" => OwnershipVoucher::deserialize_many_from_reader(&dst[..]),
```
7flying, line 62: this `add_ov`function only adds a single OV, so you should use `from_pem` and `deserialize_from_reader` instead

7flying: right, reading things again, we do want to add multiple vouchers at the same time, so the lines that say:
```
//max limit of ov size is 1024 * 32 bytes
 let mut dst = [0; 32768];
```
are misleading.

Do you mean to say that the max body size is 1024 * 32 bytes instead?

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+    if ovs.is_err() {
+        let message = ErrorCode::generate(ErrorCode::ParseError);
+        return Ok(Box::new(warp::reply::with_status(
+            warp::reply::json(&message),
+            StatusCode::BAD_REQUEST,
+        )));
+    }
+
+    // check-empty
+    let ovs = ovs.unwrap();
```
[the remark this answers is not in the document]
7flying: you can do a match over here to handle err or the result

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+    )))
+}
+
+async fn delete_ov(
+    uuids: Vec<String>,
+    user_data: Arc<OwnerServiceUD>,
+) -> core::result::Result<Box<dyn warp::reply::Reply>, warp::Rejection> {
+    let mut unknown_device = Vec::new();
+    for uuid in &uuids {
+        if Guid::from_str(uuid).is_err() {
```
7flying, line 198: since you'll be needing the unwrapped uuid also below, you can `let uuid = match [...]`  and do the error handling of the uuid first.

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+        match user_data
+            .ownership_voucher_store
+            .load_data(&Guid::from_str(uuid).unwrap())
+            .await
+        {
+            Ok(Some(_)) => {}
+            Ok(None) => {
+                unknown_device.push(uuid.to_string());
+            }
+            Err(_) => {}
```
7flying, line 214: why don't we do anything when there is an error trying to find an OV?

owner-onboarding-server/src/tls_config/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+        builder
+            .set_private_key_file(key_file, SslFiletype::PEM)
+            .unwrap();  
+        builder.build()
+    }
+
+    pub fn tls_acceptor() -> Acceptor {
+        tls_acceptor_impl(
+            "/workspaces/fido-device-onboard-rs/server.crt",
+            "/workspaces/fido-device-onboard-rs/server.key",
```
7flying, line 21: this should come from a config file

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+        ErrorCode::MiscError.to_string(),
+        StatusCode::INTERNAL_SERVER_ERROR,
+    )))
+}
+
+#[derive(Error, Debug, Serialize)]
+enum ErrorCode {
+    #[error("Voucher count should be positive integer")]
+    VoucherCountError,
+    #[error("Content-type should be application/x-pem-file,application/cbor")]
```
7flying, line 256: `application/x-pem-file,application/cbor` -> application/x-pem-file or application/cbor

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+    RequireContentTypeHeader,
+    #[error("Ownership voucher not found")]
+    NoOwnershipVoucherFound,
+    #[error("Unable to parse UUID")]
+    UuidParseError,
+    #[error("Pem format error")]
```
7flying, line 262: `Pem` -> PEM

7flying: do we always know that it is a PEM format error when we are parsing stuff?

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+                json!({"error": &ErrorCode::NoOwnershipVoucherFound.to_string()})
+            }
+            ErrorCode::UuidParseError => json!({"error": &ErrorCode::UuidParseError.to_string()}),
+            ErrorCode::StoreError(e) => json!({ "error": &e }),
+            ErrorCode::ParseError => json!(ErrorMsg {
+                error_code: "parse_error",
+                error_details: json! ({"error": ErrorCode::ParseError.to_string()})
+            }),
+            ErrorCode::InvalidNumberOfVouchers(e) => json!(ErrorMsg {
+                error_code: "invalid_number_of_voucher",
```
7flying, line 313: this should be `invalid_number_of_vouchers`

owner-onboarding-server/src/ov_management/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            ErrorCode::UnwonedVoucher(e) => json!(ErrorMsg {
+                error_code: "unowned_voucher",
+                error_details: json!({ "unowned": e })
+            }),
+            ErrorCode::UnknownDevice(e) => json!(ErrorMsg {
+                error_code: "unknown_device",
+                error_details: json!({ "unknown": e })
+            }),
+            ErrorCode::InvalidVoucherSignature(idx) => json!(ErrorMsg {
+                error_code: "invalid_voucher_signature",
```
7flying, line 325: invalid_voucher_signatures and also plural on the error code

integration-tests/tests/ov_management.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    if failed.is_empty() {
+        Ok(())
+    } else {
+        for failed_case in failed {
+            eprintln!("Failed test: {:?}", failed_case);
+        }
+        bail!("Some tests failed");
+    }
+}
```
7flying, line 71: you also need to test when a proper OV is sent in PEM format, when a proper OV is sent in cbor format, when multiple OVs are sent in both formats, when the formats are mixed up and when the actual OV count does not match the sent OV count.

owner-onboarding-server/src/ov_management/mod.rs
rdotjain: yes my bad, this is the max body size. Also, it seems like we are capping all requests at 1024*16 bytes for body size, will make this change. https://github.com/fdo-rs/fido-device-onboard-rs/blob/7ca765538d151ca6950c291a0ef4710a66e0fecd/http-wrapper/src/server.rs#L426

PR #590, opened 2023-12-14
fix: cargo test failing in dev container

### LINKED ISSUE #552: Cargo test no longer works in a clean dev environment
After commit d3d912010d7d8d41447fba10f451471e66a80c38, there's tests being executed that make sure a user exists, which isn't created in the dev container.
As a consequence, running `cargo test` there fails.

This should get fixed, to be inline with the contributing guidelines (and for easdy development for contributors).

djach7:
fixes #552


7flying (commented):
Can we add a test of some kind to make sure that we don't break the devcontainer in the future?

7flying (approved):

PR #596, opened 2024-01-03
chore: fix clippy warnings after rustc v1.75.0 update

7flying:
fix clippy warnings after rustc v1.75.0 update


nullr0ute (approved):
LGTM

PR #597, opened 2024-01-04
chore: bump warp requirement explicitly to 0.3.6

nullr0ute:
There's a pair of security issues in warp dependencies prior to 0.3.6 so bump this to at least that so we get the fixes forr
 tungstenite and rustls dependencies.


7flying (approved):
LGTM

PR #601, opened 2024-01-13
Fix hashing of Device Certificate Chain in Ownership Voucher

ben-krieger:
The specification states that “OVDevCertChainHash is the Hash of the concatenation of the contents of each byte string in OwnershipVoucher.OVDevCertChain, in the presented order."

The current behavior is to serialize the X5CHAIN to CBOR (i.e. `[ bstr[cert1] ... bstr[certN] ]` instead of `cert1 || ... || certN`).

See https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#OwnershipVoucher

BREAKING CHANGE: The device certificate chain hash in ownership voucher is now calculated from the concatenation of certificate ASN.1 DER contents rather than the CBOR serialization of an array of byte strings.

### DIFF
--- a/manufacturing-server/src/handlers/di.rs
+++ b/manufacturing-server/src/handlers/di.rs
@@ -110,7 +110,14 @@ pub(crate) async fn app_start(
     let device_certificate_chain =
         create_device_cert_chain(&user_data.device_cert_chain, device_certificate);
     let device_certificate_chain_serialized = device_certificate_chain
-        .serialize_data()
+        .chain()
+        .iter()
+        .try_fold(vec![], |mut bytes, cert| {
+            cert.to_der().map(|der| {
+                bytes.extend(der);
+                bytes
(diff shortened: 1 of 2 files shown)


ben-krieger:
Two additional points:

1. I couldn't find any code that performed device certificate chain hash verification for the ownership voucher. If I missed it, it may need updating. (And if it doesn't exist, maybe it should?)
2. This still doesn't properly cover the fact that a null certificate chain should be encoded as CBOR null. I don't think this missing case matters, though, because a null certificate chain is only for Intel EPID and similar device identities, which this library does not support.

puiterwijk (approved):
Thanks for this, this PR looks good to me.
It is quite unfortunate, but ah well.

PR #608, opened 2024-01-23
chore: bump tss-esapi crate to 7.4

nullr0ute:
There's a few minor fixes and no changes in API so let's bump to 7.4 for TPM2 support.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,7 +157,7 @@ checksum = "b9ccdd8f2a161be9bd5c023df56f1b2a0bd1d83872ae53b71a84a12c9bf6e842"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.18",
+ "syn 2.0.48",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


7flying:
I don't understand why this breaks the rust-aws-nitro-enclaves-cose too

nullr0ute:
> I don't understand why this breaks the rust-aws-nitro-enclaves-cose too

I think the breakage is the known issue we already have with rust-aws-nitro-enclaves-cose where we patched it in Fedora.

7flying:
> > I don't understand why this breaks the rust-aws-nitro-enclaves-cose too
> 
> I think the breakage is the known issue we already have with rust-aws-nitro-enclaves-cose where we patched it in Fedora.

yes, but it wasn't showing up upstream before, I need to take a look to check how we can avoid this upstream too, otherwise our tests and builds are all broken

nullr0ute:
> > > I don't understand why this breaks the rust-aws-nitro-enclaves-cose too
> > 
> > 
> > I think the breakage is the known issue we already have with rust-aws-nitro-enclaves-cose where we patched it in Fedora.
> 
> yes, but it wasn't showing up upstream before, I need to take a look to check how we can avoid this upstream too, otherwise our tests and builds are all broken

I think I can add a commit to this PR that will mitigate it

7flying (approved):
Packit builds are failing but we'll be ignoring them for the time being

PR #612, opened 2024-02-02
chore(devcontainer): setup for podman, add rust-analyzer

runcom:
People with a mac should use a podman machine that is _rootless_ and it works like a charm :)


7flying (approved):
LGTM

PR #617, opened 2024-02-09
Extend the 'per-device serviceinfo configuration' feature with files and commands

RZandvoort:
Implemented the `files` and `commands` part of the [per-device serviceinfo configuration](https://github.com/fdo-rs/fido-device-onboard-rs/blob/main/HOWTO.md#how-to-use-the-per-device-serviceinfo-feature) feature: 

## Configration example

Current example yml
```yaml
initial_user:
  username: username_per_device
  sshkeys:
  - "testkeyperdevice"
files: null
commands: null
diskencryption_clevis: null
additional_serviceinfo: null
after_onboarding_reboot: false
```

New example yml
```yaml
initial_user:
  username: username_per_device
  sshkeys:
  - "testkeyperdevice"
files:
- path: /var/lib/fdo/service-info-api/files/hosts
  permissions: 644
  source_path: /server/local/etc/hosts
commands:
- command: ls
  args:
  - /etc/hosts
  return_stdout: true
  return_stderr: true
- command: ls
  args:
  - /etc/doesnotexist/whatever.foo
  may_fail: true
  return_stdout: true
  return_stderr: true
- command: touch
  args:
  - /etc/command-testfile
diskencryption_clevis: null
additional_serviceinfo: null
after_onboarding_reboot: false

```

**Per defices files:**
It will write both base and per-device configured files.
If duplicate files are defined, it will choose the per-device file over the base file.

**Per device commands:**
Additional per device commands will be executed after the base commands


RZandvoort:
> Can we please have a description in PR, it's useful for justification etc.
> 
> Also note we're moving to DB backends so I think this would be a better fit in the DB from a layout and performance PoV.
> 
> I think also the commits could have better explanations rather than just "Update fileX". I'm going to set this to a draft.

This pull request was supposed to be for trying stuff out, but then I accidentally clicked on the big green 'create PR' button. If we eventually could squash commit, it will look like nothing happend 🤐

nullr0ute:
The DB support has now landed so it will need to be part of that. I am going to convert this to a draft, once there is DB support and everything is squashed feel free to convert it back to ready then.

RZandvoort:
I'll close this PR and will come back with a new one in the future that uses the database support.

PR #625, opened 2024-02-20
chore: update patch for new release

nullr0ute:
Update the fedora patch to fix the aws-cose crate issue with the new version.


7flying (approved):
LGTM

PR #632, opened 2024-02-28
fix(license): replace space with - in Apache 2.0

7flying:
There is an instance of "Apache 2.0" instead of "Apache-2.0" which is not allowed.


.github/commitlint.config.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable import/no-extraneous-dependencies */
-const maxLineLength = 72
+const { maxLineLength } = require('@commitlint/ensure')
```
nullr0ute, line 2: This looks like a revert on one of the other two patches.

7flying: yes, I was trying to fix that but node-js has defeated me, I've created a separate issue #633  and I'll clean this PR in a second

nullr0ute (approved):
LGTM

Issue #630, opened 2024-02-27
Tests not working on non-devcontainer environments


mmartinv:
Running tests in a non-devcontainer env with an unprivileged user still fails:
```
cargo test
   Compiling fdo-data v0.5.0 (/home/mmartinv/devel/src/github.com/fdo-rs/fido-device-onboard-rs/libfdo-data)
    Finished test [unoptimized + debuginfo] target(s) in 1.06s
     Running unittests src/main.rs (target/debug/deps/fdo_admin_tool-ff08e158d3c9431c)

running 1 test
test aio::configure::test::serialize_and_deserialize ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s

     Running unittests src/main.rs (target/debug/deps/fdo_client_linuxapp-9efbac98d63cc103)

running 4 tests
test serviceinfo::test::test_binaryfileinprogress_destination_path ... ok
test serviceinfo::test::test_pw_encryption ... ok
useradd: Permission denied.
useradd: cannot lock /etc/passwd; try again later.
test serviceinfo::test::test_user_creation_no_pw ... FAILED
useradd: Permission denied.
useradd: cannot lock /etc/passwd; try again later.
test serviceinfo::test::test_user_creation_with_pw ... FAILED

failures:

---- serviceinfo::test::test_user_creation_no_pw stdout ----
thread 'serviceinfo::test::test_user_creation_no_pw' panicked at client-linuxapp/src/serviceinfo.rs:875:9:
assertion failed: create_user(test_user).is_ok()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- serviceinfo::test::test_user_creation_with_pw stdout ----
thread 'serviceinfo::test::test_user_creation_with_pw' panicked at client-linuxapp/src/serviceinfo.rs:890:9:
assertion failed: create_user_with_password(test_user, test_password).is_ok()


failures:
    serviceinfo::test::test_user_creation_no_pw
    serviceinfo::test::test_user_creation_with_pw

test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

error: test failed, to rerun pass `-p fdo-client-linuxapp --bin fdo-client-linuxapp`
```

PR #635, opened 2024-02-28
fix: cargo test for non-root users

mmartinv:
- **fix: add `system_info` module to `utils`**
- **fix: do not run tests that require root**
- **fix: use `system_info` mod functions to gather the current user name**

Fixes: #630

### DIFF
--- a/integration-tests/tests/common/mod.rs
+++ b/integration-tests/tests/common/mod.rs
@@ -9,6 +9,8 @@ use std::{
     time::{Duration, Instant},
 };
 
+use fdo_util::system_info::get_current_user_name;
+
 use anyhow::{bail, Context as _, Result};
 use openssl::{
     asn1::{Asn1Integer, Asn1Time},
@@ -808,9 +810,7 @@ impl<'a> TestServerConfigurator<'a> {
                     &self.test_context.runner_path(&self.server_number),
                 );
 
-                let output = Command::new("whoami").output().unwrap();
-
-                let cur_user = String::from_utf8(output.stdout).unwrap();
+                let cur_user = get_current_user_name();
 
                 if !per_device {
                     L.l("per_device_serviceinfo is not set, using default values");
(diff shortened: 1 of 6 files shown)


util/Cargo.toml
```diff
@@ -19,3 +19,4 @@ fdo-http-wrapper = { path = "../http-wrapper", version = "0.5.0", features = ["s
 serde_yaml = "0.9"
 serde_cbor = "0.11"
 serde_json = "1"
+sysinfo = "0.30.5"
```
nullr0ute, line 22: Can this functionality made a dev dependency, or will they be used for things other than tests?

mmartinv: Currently this is used in tests only. Not sure if it will be used or not in other parts of the code in the future though.

mmartinv:
Opened a PR to update the [sysinfo crate](https://src.fedoraproject.org/rpms/rust-sysinfo/pull-request/3)

util/Cargo.toml
mmartinv: Actually is used in `integration-tests` which is not part of the unit tests, so I guess it cannot be moved to `[dev-dependencies]`

integration-tests/tests/common/mod.rs
```diff
@@ -808,9 +810,7 @@ impl<'a> TestServerConfigurator<'a> {
                     &self.test_context.runner_path(&self.server_number),
                 );
 
-                let output = Command::new("whoami").output().unwrap();
```
7flying, line 811: We originally had a dependency for getting the current user, and changed it to calling the `whoami` command.

I'm reluctant to adding back another dependency (with adds many other dependencies on cascade) when we can get this info without any external crate

mmartinv: Should I remove the `systeminfo` dependency then and use `whoami` in all cases?. It doesn't seem the more "elegant" way to do this but I understand your point...

miabbott: https://github.com/fdo-rs/fido-device-onboard-rs/commit/574d4e4675af3f57ba657df6ed17578e1f5305ea

:arrow_up: change from `users` crate to `whoami`

`whoami` is part of coreutils, which seems pretty safe to rely on for integration tests.

mmartinv: It looks like `whoami` has been deprecated in some distributions in favor of `id` (also in coreutils package) should I use that command instead?

miabbott: Sure; ‘id -un` operates the same as `whoami` according to the manpage

util/Cargo.toml
mmartinv: I have removed sysinfo dependency as per @7flying comments

miabbott (approved):
Nice separation of the commits; LGTM

Issue #633, opened 2024-02-28
`commitlint` is broken, CI gets stuck


7flying:
```
error running commitlint
require() of ES Module /github/workspace/node_modules/@commitlint/ensure/lib/index.js from /github/workspace/.github/commitlint.config.js not supported.
Instead change the require of index.js in /github/workspace/.github/commitlint.config.js to a dynamic import() which is available in all CommonJS modules.
Error [ERR_REQUIRE_ESM]: require() of ES Module /github/workspace/node_modules/@commitlint/ensure/lib/index.js from /github/workspace/.github/commitlint.config.js not supported.
Instead change the require of index.js in /github/workspace/.github/commitlint.config.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/github/workspace/.github/commitlint.config.js:2:27)
    at module.exports (/node_modules/import-fresh/index.js:32:59)
    at loadJsSync (/node_modules/@commitlint/load/node_modules/cosmiconfig/dist/loaders.js:[17](https://github.com/fdo-rs/fido-device-onboard-rs/actions/runs/8079062513/job/22072650131?pr=632#step:6:18):12)
    at loadJs (/node_modules/@commitlint/load/node_modules/cosmicon
```

apparently 

```js
const { maxLineLength } = require('@commitlint/ensure')
```
in our `.github/commitlint.config.js` is no longer allowed.

PR #636, opened 2024-02-28
chore: fix require error with commitlint

miabbott:
Changed the commitlint config file to use a dynamic import of `commitlint/esnure` and updated the function to await the promise of the import.

(This code was generated by ChatGPT and appears to fix the error with using `require('@commitlint/ensure')`)

Closes: #633


nullr0ute (approved):
LGTM

Issue #642, opened 2024-03-06
Add a `sqlite` config option to the `aio` option of `fdo-admin-cli`


7flying:
The user should be able to specify that they want a `sqlite` database as the storage method when running the servers with `fdo-admin-cli aio`

The user should still manually export the OVs from the manufacturing server's DB and then import them into the owner server's DB.

nullr0ute:
Maybe we should add a "backend" option and have sqlite/files as the options and default to sqlite

7flying:
Once sqlite has been enabled in `aio`, other features that depend on database support, such as the re-registration window of OVs in the owner server should also be configured.

PR #643, opened 2024-03-08
Add an OV re-registration window option when using DB storage

### LINKED ISSUE #193: Owner Onboarding Server: make sure to re-run TO0 before it is actually expired

7flying:
This allows the users to specify an OV re-registration window in the Owner server. When an OV is about to be de-registered in the rendezvous server within the specified time window, the owner will trigger a re-registration.

Fixes #193 
Obsoletes #207

### DIFF
--- a/admin-tool/src/aio/configure.rs
+++ b/admin-tool/src/aio/configure.rs
@@ -7,6 +7,8 @@ use std::{collections::BTreeMap, net::IpAddr, path::Path};
 
 use fdo_store::StoreConfig;
 use fdo_util::servers::configuration::{
+    owner_onboarding_server::DEFAULT_REGISTRATION_PERIOD,
+    owner_onboarding_server::DEFAULT_RE_REGISTRATION_WINDOW,
     serviceinfo_api_server::ServiceInfoSettings, AbsolutePathBuf, Bind,
 };
 
@@ -46,6 +48,11 @@ pub(super) struct Configuration {
     #[clap(long)]
     pub manufacturing_use_secp256r1: bool,
 
+    #[clap(long)]
+    ov_registration_period: Option<u32>,
+    #[clap(long)]
+    ov_re_registration_window: Option<u32>,
+
     /// The hostname or IP address that clients should use to connect to the AIO components
     /// (if not specified, will be all IP addresses of the system).
     /// Note that this is not equal to the listen address, as the AIO components will always
@@ -57,6 +64,7 @@ pub(super) struct Configuration {
     // and generate the intermediate data
     #[clap(skip)]
     pub contact_addresses: Vec<ContactAddress>,
+
     #[clap(skip)]
     pub serviceinfo_api_auth_token: String,
     #[clap(skip)]
@@ -87,6 +95,9 @@ impl Default for Configuration {
             manufacturing_disable_key_storage_filesystem: false,
             manufacturing_disable_key_storage_tpm: true,
             manufacturing_use_secp256r1: false,
+
+            ov_registration_period: Some(DEFAULT_REGISTRATION_PERIOD),
+            ov_re_registration_window: Some(DEFAULT_RE_REGISTRATION_WINDOW),
 
             contact_hostname: None,
 
@@ -339,6 +350,9 @@ fn generate_configs(aio_dir: &Path, config_args: &Configuration) -> Result<(), E
                 .generate_owner_addresses()
                 .context("Error generating owner addresses")?,
             report_to_rendezvous_endpoint_enabled: true,
+
+            ov_registration_period: config_args.ov_registration_period,
+            ov_re_registration_window: config_args.ov_re_registration_window,
         };
     write_config(
         aio_dir,
--- a/owner-onboarding-server/src/main.rs
+++ b/owner-onboarding-server/src/main.rs
@@ -27,9 +27,14 @@ use fdo_data_formats::{
     publickey::PublicKey,
     types::{Guid, TO2AddressEntry},
 };
-use fdo_store::{Store, StoreError};
+
+use fdo_store::{Store, StoreConfig, StoreError};
 use fdo_util::servers::{
-    configuration::{owner_onboarding_server::OwnerOnboardingServerSettings, AbsolutePathBuf},
+    configuration::{
+        owner_onboarding_server::OwnerOnboardingServerSettings,
+        owner_onboarding_server::DEFAULT_REGISTRATION_PERIOD,
+        owner_onboarding_server::DEFAULT_RE_REGISTRATION_WINDOW, AbsolutePathBuf,
+    },
     settings_for, OwnershipVoucherStoreMetadataKey,
 };
 
@@ -63,6 +68,13 @@ pub(crate) struct OwnerServiceUD {
     service_info_api_client: fdo_http_wrapper::client::JsonClient,
 
     owner_addresses: Vec<TO2AddressEntry>,
+
+    // How much time (s) OVs are going to be registered
+    ov_registration_period: u32,
+    // The time window (s) within which the re-registration will start
+    ov_re_registration_window: u32,
+
+    window_check_enabled: bool,
 }
 
 pub(crate) type OwnerServiceUDT = Arc<OwnerServiceUD>;
@@ -73,7 +85,14 @@ fn load_private_key(path: &AbsolutePathBuf) -> Result<PKey<Private>> {
 }
 
 async fn _handle_report_to_rendezvous(udt: &OwnerServiceUDT, ov: &OwnershipVoucher) -> Result<()> {
-    match report_ov_to_rendezvous(ov, &udt.owner_addresses, &udt.owner_key).await {
+    match report_ov_to_rendezvous(
+        ov,
+        &udt.owner_addresses,
+        &udt.owner_key,
+        udt.ov_registration_period,
+    )
+    .await
+    {
         Ok(wait_seconds) => {
             udt.ownership_voucher_store
                 .store_metadata(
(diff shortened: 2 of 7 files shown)


admin-tool/src/aio/configure.rs
```diff
@@ -46,6 +46,11 @@ pub(super) struct Configuration {
     #[clap(long)]
     pub manufacturing_use_secp256r1: bool,
 
+    #[clap(long)]
```
djach7, line 49: Just a question: Is this how the clap cli parser is implemented? I've read a little about it a while ago but haven't seen it used before.

owner-onboarding-server/src/main.rs
```diff
@@ -63,6 +63,13 @@ pub(crate) struct OwnerServiceUD {
     service_info_api_client: fdo_http_wrapper::client::JsonClient,
 
     owner_addresses: Vec<TO2AddressEntry>,
+
+    // How much time (s) OVs are going to be registered
+    ov_registration_period: u32,
+    // The time window (s) within when the re-registration will start
```
djach7, line 69: Nit: "The time window (s) within _which_ the re-registration will start"

owner-onboarding-server/src/main.rs
```diff
@@ -138,10 +152,51 @@ async fn report_to_rendezvous(udt: OwnerServiceUDT) -> Result<()> {
     Ok(())
 }
 
+async fn check_registration_window(udt: &OwnerServiceUDT) -> Result<()> {
+    let now_plus_window =
+        time::OffsetDateTime::now_utc().unix_timestamp() + (udt.ov_re_registration_window as i64);
+    // these are the ovs whose registration time will end and we need to
```
djach7, line 158: Extra Nitty: It seems like all other comments are capitalized, does it make sense to capitalize these too?

djach7 (dismissed):
LGTM, reviewed for English conventions with limited rust knowledge.

owner-onboarding-server/src/main.rs
7flying: I usually capitalise when I'm describing a function or global parameters or struct fields, but I don't / don't care when  I'm making a comment inside a function

admin-tool/src/aio/configure.rs
7flying: yes, if you look at the top of the struct there are a couple of `Derive`s:
```rust
#[derive(Debug, Args, Clone, Serialize, Deserialize)]
```
 `Debug` and `Clone` are Rust-specific derives, `Serialize` and `Deserialize` are from Serde (see the imports) and `Args` comes from Clap (see the import).
With Clap you can use Derives to define arguments to the cli functions that you will be implementing if you define in the Cargo.toml that you want the 'Derive' feature (see: https://github.com/fdo-rs/fido-device-onboard-rs/blob/main/admin-tool/Cargo.toml#L15). Then, with the `Args` derive you are telling Clap that those are arguments to cli function, and with the specific `#[clap(long)]`, `#[clap(whaetever)]` you're telling Clap how to parse the options.
There is a small example on how to add a new cli function here: https://github.com/fdo-rs/fido-device-onboard-rs/commit/4da0a49bf9d7f272b86b07e009b0dde3bfc5600d
And a more complete example on how you would implement a full cli with Clap from scratch (`manufacturing-client/src/main.rs` is the relevant file): https://github.com/fdo-rs/fido-device-onboard-rs/commit/dfe51b2803d2a8aa47ef939130c7391426def1e6

owner-onboarding-server/src/main.rs
```diff
@@ -242,14 +296,21 @@ async fn perform_maintenance(udt: OwnerServiceUDT) -> std::result::Result<(), &'
         #[allow(unused_must_use)]
         let (ov_res, ses_res, rtr_res) = tokio::join!(ov_maint, ses_maint, rtr_maint);
 
+        if udt.window_check_enabled {
+            let window_res = check_registration_window(&udt.clone()).await;
+            if let Err(e) = window_res {
+                log::warn!("Error during re-registration window check: {e:?}");
+            }
+        }
+
```
mmartinv, line 305: If `ov_re_registration_window` is less than `MAINTENANCE_INTERVAL` the OVs on Rendezvous server will be expired before the `check_registration_window` function is called again.

Although the expired OVs will be registered again by the `report_to_rendezvous` function after `MAINTENANCE_INTERVAL - ov_re_registration_window`  seconds the purpose of this PR, which I guess is to prevent the OV registration to expire on Rendezvous server side, is missed, right?.

Maybe `MAINTENANCE_INTERVAL` should be adjusted by taking into account the value of `ov_re_registration_window` or just check the registration window in a separate thread maybe?

mmartinv (commented):
Having to implement `query_ovs_db_to2_performed_to0_less_than` function 6 times in which only 2 of them are "real" implementations and 4 just throw an error feels very wrong to me. Imagine we need to implement another database backend (e.g. mysql), now we have 9 implementations: 3 real, 6 throwing errors. I think we need to revisit the whole thing sooner than later.

7flying:
> Having to implement `query_ovs_db_to2_performed_to0_less_than` function 6 times in which only 2 of them are "real" implementations and 4 just throw an error feels very wrong to me. Imagine we need to implement another database backend (e.g. mysql), now we have 9 implementations: 3 real, 6 throwing errors. I think we need to revisit the whole thing sooner than later.

I do agree, we should get rid of the `directory` storage method, which is the one that defines the trait that we need to implement and doesn't fit at all with the DB stuff.

admin-tool/src/aio/configure.rs
```diff
@@ -88,6 +94,9 @@ impl Default for Configuration {
             manufacturing_disable_key_storage_tpm: true,
             manufacturing_use_secp256r1: false,
 
+            ov_registration_period: Some(600),
+            ov_re_registration_window: Some(100),
```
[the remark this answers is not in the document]
7flying: I wonder whether we should define a new fdo-constants or fdo-defaults for that since it doesn't look right to me to put those constants in the fdo-data-formats crate. what do you think?

admin-tool/src/aio/configure.rs
```diff
@@ -88,6 +94,9 @@ impl Default for Configuration {
             manufacturing_disable_key_storage_tpm: true,
             manufacturing_use_secp256r1: false,
 
+            ov_registration_period: Some(600),
```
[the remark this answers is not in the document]
7flying: see below

admin-tool/src/aio/configure.rs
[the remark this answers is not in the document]
mmartinv: I think the best fit would be `util/src/servers/configuration/owner_onboarding_server.rs`. Not sure if putting constants there is an antipattern in rust or goes against something though.

owner-onboarding-server/src/main.rs
```diff
@@ -138,10 +157,51 @@ async fn report_to_rendezvous(udt: OwnerServiceUDT) -> Result<()> {
...
+        .await
+        {
+            Ok(wait_seconds) => {
+                udt.ownership_voucher_store
+                    .store_metadata(
+                        ov.header().guid(),
+                        &fdo_store::MetadataKey::Local(
+                            OwnershipVoucherStoreMetadataKey::To0AcceptOwnerWaitSeconds,
+                        ),
+                        &time::Duration::new(wait_seconds.into(), 0),
```
mmartinv, line 185: The rendezvous server may return a `wait_seconds` which is less than `ov_registration_period`. Maybe we should also check that is also bigger than `ov_re_registration_window` and  print a warning in the log saying that the re registration will not work for that ownership voucher and/or maybe disable `window_check_enabled`?

7flying: Good point will add the warning. The `wait_seconds` that comes from the redezvous should be what the owner is requesting, _or less_, the only config that the rendezvous has is an upper limit, that we cannot know (this is just in our implementation).

mmartinv (approved):
Looks good to me. Just a nitpick: you marked a conversation as resolved after an ack but you didn't take action on it. If you are not willing to implement the suggested change because it was described as an enhancement and you considered it as not important just say it, do not just resolve the conversation without further explanation, it feels awkard.

7flying:
> Looks good to me. Just a nitpick: you marked a conversation as resolved after an ack but you didn't take action on it. If you are not willing to implement the suggested change because it was described as an enhancement and you considered it as not important just say it, do not just resolve the conversation without further explanation, it feels awkard.

I missed that, let me review and create an issue so that it is not forgotten

owner-onboarding-server/src/main.rs
```diff
@@ -335,6 +401,41 @@ async fn main() -> Result<()> {
...
+        }
+    };
+
+    if ov_re_registration_window >= ov_registration_period {
+        bail!(
+            "ov_re_registration_window ({ov_re_registration_window}) must be smaller than ov_registration_period ({ov_registration_period})");
+    } else {
+        log::info!("Server configured with an OV registration period of {ov_registration_period} seconds, OV re-registration window set to {ov_re_registration_window} seconds")
+    }
+
```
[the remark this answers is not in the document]
7flying: @mmartinv I'm guessing that this is it, let me explain since I did not comment.
I have a pending issue to add sqlite support as an optional argument when running aio (#642 ), as part of that I would have to change the passed arguments to the aio command, and it makes sense to add a way to configure other things (such as this new registration window there), so, yeah I made a mental note of this but did not comment. My apologies.

7flying:
> > Looks good to me. Just a nitpick: you marked a conversation as resolved after an ack but you didn't take action on it. If you are not willing to implement the suggested change because it was described as an enhancement and you considered it as not important just say it, do not just resolve the conversation without further explanation, it feels awkard.
> 
> I missed that, let me review and create an issue so that it is not forgotten

OK, commented in the review since I missed to comment what I was thinking

PR #648, opened 2024-03-15
feat: use vendored sources

mmartinv:
- **fix: use official `aws-nitro-enclaves-cose` crate**
- **fix: do not generate vendor tarball in builds**
- **fix: lock clap_builder to 4.4 version**
- **feat: use patched vendored sources for builds.**
- **fix: update packit configuration**


fido-device-onboard.spec
```diff
@@ -35,19 +33,18 @@ BuildRequires:  tpm2-tss-devel
...
 %cargo_prep -v vendor
 %else
-%cargo_prep -V 1
+%cargo_prep -V 0
 %endif
 %endif
 
 %if 0%{?fedora}
-%autosetup -p1 -n %{name}-rs-%{version}
+rm -rf .cargo vendor
```
nullr0ute, line 47: We don't use vendoring in Fedora.

fido-device-onboard.spec
```diff
@@ -35,19 +33,18 @@ BuildRequires:  tpm2-tss-devel
...
+%autosetup -p1 -n fido-device-onboard-rs-0.5.0
 
 %if 0%{?rhel}
-%autosetup -p1 -a1 -n %{name}-rs-%{version}
-rm -f Cargo.lock
 %if 0%{?rhel} >= 10
 %cargo_prep -v vendor
 %else
-%cargo_prep -V 1
+%cargo_prep -V 0
```
nullr0ute, line 42: Generates the vendored source at rpm build time? In the case of RHEL/c9s/eln/Fedora that is not going to work as they don't have out bound connectivity. For copr builds it depends on the repo config but it should be fine for CI builds in copr.

.packit.yaml
```diff
@@ -1,48 +1,50 @@
...
+      job: copr_build
       trigger: pull_request
       targets:
-          - fedora-development-aarch64
           - fedora-development
+          - fedora-development-aarch64
           - fedora-latest
           - fedora-latest-aarch64
-    - job: copr_build
+          - centos-stream-9-x86_64
```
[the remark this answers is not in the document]
miabbott: Looks like we might be able to do EPEL builds for cs9?  re:  https://github.com/containers/podman/blob/main/.packit.yaml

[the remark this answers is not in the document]
nullr0ute: We could do scratch EPEL builds, but the EPEL rules are that packages can't duplicate what's in RHEL so we can't package it there because it'll conflict and customers end up in a mess!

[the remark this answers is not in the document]
miabbott: The rule makes sense, though I wonder how `podman` is getting around it.

For our purposes, it seems like skipping c9s and EPEL would be wise.

fido-device-onboard.spec
mmartinv: From my tests it looks like the `-V 1` specifies the source tarball containing the vendor files. Now those files are contained in the `Source0` tarball and actually works

fido-device-onboard.spec
```diff
@@ -312,6 +309,28 @@ Requires: fdo-init = %{version}-%{release}
 %systemd_postun_with_restart fdo-aio.service
 
 %changelog
+* Fri Mar 15 2024 Miguel Martín <mmartinv@redhat.com> - 0.5.0-1.20240315124438915936.use.vendored.sources.28.gd33cb89d
+- fix: update packit configuration (Miguel Martín)
```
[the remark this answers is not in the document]
mmartinv: Yeah that should not be there :)

data-formats/Cargo.toml
```diff
@@ -17,7 +17,7 @@ serde_cbor = "0.11"
 serde_repr = "0.1.6"
 serde_tuple = "0.5"
 thiserror = "1"
-aws-nitro-enclaves-cose = { git = "https://github.com/nullr0ute/aws-nitro-enclaves-cose/", rev = "e3938e60d9051690569d1e4fcbe1c0c99d2fafa8" }
+aws-nitro-enclaves-cose = "0.4.0"
```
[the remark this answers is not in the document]
mmartinv: when running `update-vendor.sh` script

.packit.yaml
```diff
@@ -1,48 +1,50 @@
...
-
 jobs:
-    - job: copr_build
+    - &fdo_copr
+      job: copr_build
       trigger: pull_request
       targets:
-          - fedora-development-aarch64
           - fedora-development
+          - fedora-development-aarch64
```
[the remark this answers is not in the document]
mmartinv: Unfortunately for packit the absence of arch means `x86_64` only so we need the arch specific suffix.

.packit.yaml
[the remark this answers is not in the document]
mmartinv: I am not sure I am following... The centos-stream9 builds are just builds triggered in packit's copr not Koji or any other downstream copr/dist-git or whatever so I don't think we need any commit formats/approved bugs or anything else. The packit produced RPMs will be useful to run the tests on the specific platforms and that what we needed, right?.

On the other hand, I might be wrong but the EPEL builds @miabbott is referring to are just copr builds of CentOS Stream 9 with the EPEL-next repo activated, nothing more, the resulting RPMs are not going to be in EPEL-next, they just need EPEL-next to be built. As we don't need any EPEL repository to build the FDO packages I will leave centos-stream9 as it is.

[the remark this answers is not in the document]
miabbott: TBH, I'm not terribly familiar with how Packit handles c9s, so I was just looking for examples of other projects using it.

To that point, `osbuild` is using it for all downstream builds - https://github.com/osbuild/osbuild-composer/blob/main/.packit.yaml

And looking at their dashboard from Packit, shows that the c9s builds are just going to a copr - https://dashboard.packit.dev/projects/github.com/osbuild/osbuild

So I'll walk back my statement about skipping c9s...seems like it is something we can try.  (We'll still need to do manual work to get "official" c9s builds made...or invest in a bot to do it for us like the `osbuild` folks have)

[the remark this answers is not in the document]
nullr0ute: > I am not sure I am following... The centos-stream9 builds are just builds triggered in packit's copr not Koji or any 

So by centos-stream9 builds you're actually referring to copr builds against the c9s? I thought you meant as in c9s proper, when read "centos-stream9 builds" I read it as official c9s builds not copr el9 builds. Anything in copr is fine.

fido-device-onboard.spec
nullr0ute: I don't believe we want it all as single tarball.

mmartinv: Could you please elaborate further on why we wouldn't want a single tarball?

7flying: We don't build rhel/c9s using the Makefile, we need an srpm and the vendored tarfile with the dependencies, which is uploaded to the build system.

7flying:
I think that we are trying to accomplish too many things with these PR when we don't really need all those things. Here are my thoughts and what I understand to be "true"

First of all, as I understand it, this project simply needs to check that things work in Fedora, thereby, as it was before, we only care about making copr builds in Fedora rawhide and the latest Fedora development, both for aarch64 and x86_64.

We don't need to test that packages are built in rhel/c9s because we vendor everything, and since we compile everything ourselves """It Just Works""". It would be nice to have a rhel/c9s CI but as I've understood the PR, it does not replicate the way that we need to build things for rhel/c9s, so if it does not replicate it I would rather save time in CI and I wouldn't run rhel/c9s packaging tests. 

When things do not work in rhel/c9s we make specific patches for it in the specific c9s repo, but for "upstream-sake" we had the make-vendored-tarfile script over here too. Due to the way that packaging works in rhel/c9s we do not need a Makefile that spits out an RPM, we need a sprm and a separate tarball with the vendored dependencies.
The only thing that I need is a better make-vendored-tarfile or X that filters the windows stuff better.

.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[source.crates-io]
+replace-with = "vendored-sources"
+
+[source.vendored-sources]
+directory = "vendor"
```
7flying, line 5: this is going to use the vendored sourcesfor everyt build and we don't want that

fido-device-onboard.spec
nullr0ute: Because that would mean we ship a vendored tarball in Fedora which we don't want and it's the way it's generally done overall within specs. Also it keeps out source separate from the third party stack.

.cargo/config.toml
mmartinv: Well that's not actually true, the `%cargo_prep` macro overwrites it. Only when `%cargo_prep -V 0` or `%cargo_prep -v vendor`  the build will use vendored sources.

fido-device-onboard.spec
mmartinv: We still need the vendored tarball in fedora for ELN builds don't we? and ELN is built from Fedora `dist-git` right? So we need the vendored tarball in all the distributions we build for, don't we?. What's the difference on shipping it in a separate tarball than alongside the source code? Other projects already do that like all the osbuild projects for example.

nullr0ute: What do the Fedora rust packaging guidelines say? https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/

I would personally like them to be separate.

mmartinv:
> I think that we are trying to accomplish too many things with these PR when we don't really need all those things. Here are my thoughts and what I understand to be "true"
> 
> First of all, as I understand it, this project simply needs to check that things work in Fedora, thereby, as it was before, we only care about making copr builds in Fedora rawhide and the latest Fedora development, both for aarch64 and x86_64.
> 

As we discussed in the refinement session we also need to enable CentOS Stream 9 builds at least.

> We don't need to test that packages are built in rhel/c9s because we vendor everything, and since we compile everything ourselves """It Just Works""". It would be nice to have a rhel/c9s CI but as I've understood the PR, it does not replicate the way that we need to build things for rhel/c9s, so if it does not replicate it I would rather save time in CI and I wouldn't run rhel/c9s packaging tests.
> 

We need to test the packaging itself and that the resulting RPMs work as expected. The way the RPMs are built in copr is basically the same as in dist-git I believe. Indeed, packit can propose downstream dist-git updates when a new version is released upstream so I think we could even automate some downstream work.

> When things do not work in rhel/c9s we make specific patches for it in the specific c9s repo, but for "upstream-sake" we had the make-vendored-tarfile script over here too. Due to the way that packaging works in rhel/c9s we do not need a Makefile that spits out an RPM, we need a sprm and a separate tarball with the vendored dependencies. The only thing that I need is a better make-vendored-tarfile or X that filters the windows stuff better.

What's the difference compared to provide everything, the source code alongside the filtered and patched vendor files, in a single SRPM? Is there any problem I am not aware of? AFAIK osbuild guys already do that with osbuild-composer for example.

fido-device-onboard.spec
mmartinv: It literally says:

> [In general](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling), packages SHOULD NOT use bundled crate dependencies, whenever possible.

So we are OK in Fedora as we don't use vendoring at all.

nullr0ute: So using a separate upstream and vendored sourced allows me to:

- Verify the upstream source hasn't changed from the github releases download to verify checksum etc
- Refresh the vendored source against an upstream release without having to change the upstream release source to deal with CVEs in the vendored source.

Having them separate is useful for a number of reasons. Please just update this to keep them separate.

mmartinv: > So using a separate upstream and vendored sourced allows me to:
> 
>* Verify the upstream source hasn't changed from the github releases download to verify checksum etc
>

Sorry but I don't get what's the difference. If you include the `vendor` directory within the release you still can verify the checksum exactly the same way.
 
>* Refresh the vendored source against an upstream release without having to change the upstream release source to deal with CVEs in the vendored source.
> 

Applying a CVE downstream without having to modify the upstream repo would be as simple as:
```console
git clone https://github.com/fdo-rs/fido-device-onboard-rs.git
cd fido-device-onboard-rs
git checkout v0.5.0
# do the required modifications to address the CVE by changing dependencies and/or adding patches to the `patches` directory
./update-vendor.sh
git diff > cve-number.diff 
```
And use that `cve-number.diff` as patch in the dowstream repo.

> 
> Having them separate is useful for a number of reasons. Please just update this to keep them separate.

Sorry but I cannot update the PR to separate them, that's the whole point of this PR: to minimize hacks and to bring the filtered and patched vendor directory into the repo's source code. The only reason I see for now to not merge it is the problem I found regarding `cargo add` that I mentioned below. However that would be a minor problem that can be solved by:
```
rm -f .cargo/config.toml
cargo add ...
./update-vendor.sh
```

mmartinv:
Closing as this is not going to be merged

Issue #628, opened 2024-02-22
Rework trusted manufacturer checks.


mmartinv:
Currently the Rendevouz server gets the list of the trusted manufacturer CAs from the **trusted_manufacturer_keys_path** variable within the config file and [checks](https://github.com/fdo-rs/fido-device-onboard-rs/blob/main/rendezvous-server/src/handlers_to0.rs#L72-L91) if the OwnershipVoucher's Manufacturer Public Key (OwnershipVoucher.OVHeader.OVPubkey) is in the list of of the trusted manufacturer CA certificates by extracting the pub key of each CA.

But according to the [FIDO specification](https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#OwnershipVoucher) the **OwnershipVoucher.OVHeader.OVPubkey** is used for the verification of the first entry of the OwnershipVoucher.OVEntryArray only because it's just a public key, not a certificate:

> OVEPubKey is the public key that verifies the signature on the next entry’s COSE Sig1 object. The first entry is verified by **OVHeader.OVPubKey**. This creates a signature chain from the Ownership Voucher header through each entry, to the last entry. The last entry’s public key verifies a signature created during the TO2 protocol, in the TO2.ProveOVHdr message. 

To decide which manufacturer's CAs are trusted by Rendevouz server I think we should perform a verification of the **OwnershipVoucher.OVDevCertChain** instead, the same check as the [Owner](https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#owner-verification-of-device-certificate-chain) is supposed to do.

Also, it might be convenient to explore the possibility of adding support for disabling this check so the servers can trust any manufacturer.

(It seems that currently the RendeVouzServer already checks the **OwnershipVoucher.OVDevCertChain**  and [trusts all the certificates](https://github.com/fdo-rs/fido-device-onboard-rs/blob/main/rendezvous-server/src/handlers_to0.rs#L160-L161))

mmartinv:
Also from the FIDO Specification (thanks @7flying for the heads up):

[...]

PR #656, opened 2024-04-02
feat: verify trusted manufacturers

mmartinv:
- **feat(rendezvous-server)**: verify the device certificate chain
- **feat(owner-onboarding-server)**:  verify the device certificate chain
- **fix(owner-onboarding-server)**: spelling typos

Fixes: #628

### DIFF
--- a/rendezvous-server/src/handlers_to0.rs
+++ b/rendezvous-server/src/handlers_to0.rs
@@ -157,8 +157,14 @@ pub(super) async fn ownersign(
         }
         Some(v) => v,
     };
-    //let device_pubkey = match device_cert_chain.verify_from_x5bag(&user_data.trusted_device_keys) {
-    let device_pubkey = match device_cert_chain.insecure_verify_without_root_verification() {
+
+    let verify_device_pubkey = if let Some(trusted_device_keys) = &user_data.trusted_device_keys {
+        device_cert_chain.verify_from_x5bag(trusted_device_keys)
+    } else {
+        device_cert_chain.insecure_verify_without_root_verification()
+    };
+
+    let device_pubkey = match verify_device_pubkey {
         Err(cert_chain_err) => {
             log::debug!("Error verifying device certificate: {:?}", cert_chain_err);
             return Err(Error::new(
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -491,8 +491,8 @@ Where:
       Please refer to the [Database management section](#database-management) on how to initialize databases.
 - `session_store_driver`: path to a directory that will hold session
   information.
-- `trusted_device_keys_path`: path to the Device Certificate Authority
-  certificate.
+- `trusted_device_keys_path` [OPTIONAL]: path to the CA certificates
+used for device certificate chain verification.
 - `owner_private_key_path`: path to the Owner's private key.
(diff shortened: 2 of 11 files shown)


rendezvous-server/src/handlers_to0.rs
```diff
@@ -69,27 +69,6 @@ pub(super) async fn ownersign(
...
-        if !trusted_manufacturer_keys.contains_publickey(&manufacturer_pubkey) {
-            return Err(Error::new(
-                ErrorCode::InvalidOwnershipVoucher,
-                messages::v11::to0::OwnerSign::message_type(),
-                "Ownership voucher manufacturer not trusted",
-            )
-            .into());
-        }
-    }
-
```
7flying, line 92: This part is actually OK and we shouldn't delete it.
`to0d.OV.header[4]` holds the manufacturer's public key, as specified at <https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#OwnershipVoucher>.
The manufacturer server stores its public cert there ([code](https://github.com/fdo-rs/fido-device-onboard-rs/blob/main/manufacturing-server/src/handlers/di.rs#L133)), so in this part we are just doing an allowlist check.

7flying: this also comes [from this](https://fidoalliance.org/specs/FDO/FIDO-Device-Onboard-PS-v1.1-20220419/FIDO-Device-Onboard-PS-v1.1-20220419.html#TO0:~:text=It%20is%20preferred,such%20a%20trust.) :
> It is preferred that the Rendezvous Server has a basis on which to trust at least one public key within the Ownership Voucher. For example, the manufacturer who ran the DI protocol to configure the Device, thereby choosing the Rendezvous Server, may register public keys with the Rendezvous Server to establish such a trust.

rendezvous-server/src/handlers_to0.rs
```diff
@@ -157,8 +136,15 @@ pub(super) async fn ownersign(
         }
         Some(v) => v,
     };
-    //let device_pubkey = match device_cert_chain.verify_from_x5bag(&user_data.trusted_device_keys) {
-    let device_pubkey = match device_cert_chain.insecure_verify_without_root_verification() {
+
+    let device_pubkey_verification =
+        if let Some(trusted_manufacturer_certs) = &user_data.trusted_manufacturer_keys {
+            device_cert_chain.verify_from_x5bag(trusted_manufacturer_certs)
```
7flying, line 142: So, I think that the issue here is whether we can trust any certificate as root or not and whether the rendezvous has the abilities to verify root certificates...

7flying: what was the error that we were getting before changing the ca cert in the tests?

mmartinv: > what was the error that we were getting before changing the ca cert in the tests?

Honestly, I don't remember, I would need to check the conformance tests again. 
Either way, we need the servers to be able to validate the device certificate chain by using a list of trusted CA certs (apart from verifying that the chain is correct).

7flying (approved):

PR #666, opened 2024-08-14
database enhancements

runcom:
[chore: store: db: reuse connection pool](https://github.com/fdo-rs/fido-device-onboard-rs/commit/68a7ac15131028f46089d5b67e5ec91ed1a34161)

The pool object is meant to be reused and since we're initializing the
store as part of a long running process, we should save the pool object
and pass it around/use it.

[chore: store: db: refactor database configuration](https://github.com/fdo-rs/fido-device-onboard-rs/commit/b61ec7e83291ee35db04e2ca846e6084130b0e3d)

This mainly comes from @mmartinv comment https://github.com/fdo-rs/fido-device-onboard-rs/pull/556/files#r1465996442


nullr0ute:
Is this ready for review? Looking to be ongoing updates so maybe should be a draft until it's ready

7flying (commented):
You should move the removal of store/src/db.rs into the first commit since you are already creating two separate files for sqlite and postgres that have the pool reuse modification

runcom:
> You should move the removal of store/src/db.rs into the first commit since you are already creating two separate files for sqlite and postgres that have the pool reuse modification

I figured it's probably better to have this enhancement as a single commit (too many moving pieces and I want to avoid to break bisect too)

7flying:
> > You should move the removal of store/src/db.rs into the first commit since you are already creating two separate files for sqlite and postgres that have the pool reuse modification
> 
> I figured it's probably better to have this enhancement as a single commit (too many moving pieces and I want to avoid to break bisect too)

That also works for me :+1:

7flying (approved):
LGTM

PR #669, opened 2024-08-27
chore: update diesel to 2.2.3

7flying:
This closes a high severity security advisory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,7 +387,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b922faaf31122819ec80c4047cc684c6979a087366c069611e33649bf98e18d"
 dependencies = [
  "clap 3.2.25",
- "heck",
+ "heck 0.4.1",
  "indexmap 1.9.3",
  "log",
  "proc-macro2",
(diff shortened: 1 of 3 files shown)


7flying:
Fedora still hasn't 2.23 https://packages.fedoraproject.org/pkgs/rust-diesel/rust-diesel-devel/

nullr0ute (approved):
Looks good, and Fedora is up2date

mmartinv:
This change broke the CentOS Stream 9 builds as diesel 2.2 requires rust 1.78 (vs 1.75 present in CentOS stream 9)

runcom:
@mmartinv can you revert this while we figure out how to update the toolchain in centos (if we can)

nullr0ute:
are you sure it's not being bumped for the next el9 release? Do we ship vulverable stacks while awaiting catch up? I don't think this should be reverted

runcom:
I don't understand, but anyway, this was open since august and as it merged centos broke so what's the difference in reverting for a day while we figure out stuff

mmartinv:
It looks like CentOS Stream 9 is going to ship rust 1.79, not sure how soon though

nullr0ute:
> I don't understand, but anyway, this was open since august and as it merged centos broke so what's the difference in reverting for a day while we figure out stuff

I was looking to cut 5.1 to put into Fedora for fixes there as it needs the newer diesel as the older one has been dropped due to the CVEs.

nullr0ute:
> It looks like CentOS Stream 9 is going to ship rust 1.79, not sure how soon though

Looks like c10s is going to that rev too.

PR #673, opened 2024-09-11
feat(manufacturing-server): implement an export OVs endpoint 

runcom:
Just serving an archive with all the OVs the manufacturer knows about.
It'd be handy to just give this away to whoever needs these credentials
and/or create a nice UI where you click a button to have them all.
The post-MVP, with a UI, would be to just have a UI that is able to list
all the device credentials, let you select which one you want, download
them in an archive, profit. Not there yet.

This is based on #671 - so let's merge that first

Signed-off-by: Antonio Murdaca <amurdaca@redhat.com>

### DIFF
--- a/manufacturing-server/src/main.rs
+++ b/manufacturing-server/src/main.rs
@@ -1,8 +1,14 @@
 use std::collections::BTreeMap;
 use std::convert::{TryFrom, TryInto};
-use std::fs;
+use std::fs::{self, File};
+use std::io::Read;
 use std::str::FromStr;
 use std::sync::Arc;
+
+use fdo_data_formats::{constants::ErrorCode, ProtocolVersion};
+use fdo_store::Store;
+
+use warp::{Filter, Rejection};
 
 use anyhow::{bail, Context, Error, Result};
 use openssl::{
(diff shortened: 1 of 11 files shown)


store/src/sqlite.rs
```diff
@@ -461,6 +497,24 @@ where
     V: Serializable + Send + Sync + Clone + 'static,
     MKT: crate::MetadataLocalKey + 'static,
 {
+    async fn load_all_data(&self) -> Result<Vec<V>, StoreError> {
```
say-paul, line 500: Why not referring a single function for load_all_data?

manufacturing-server/src/main.rs
```diff
@@ -268,7 +274,49 @@ async fn main() -> Result<()> {
...
+                }
+            },
+        ))
+        .map(|ovs: Vec<OwnershipVoucher>| {
+            let tmp_dir = TempDir::new("manufacturer-server-ovs").unwrap();
+            for ov in ovs {
+                let file_path = tmp_dir.path().join(ov.header().guid().to_string());
+                let tmp_file = File::create(file_path).unwrap();
+                OwnershipVoucher::serialize_to_writer(&ov, &tmp_file).unwrap();
+            }
```
mmartinv, line 297: Shouldn't we return 404 if there are no OVs?

mmartinv (approved):
I have tested the changes and other than the comment above, it looks good to me.

store/src/sqlite.rs
runcom: rust won't allow that across different implementation of the store type (or it does using generics but I'm not there yet :))

manufacturing-server/src/main.rs
runcom: do you think it's better a 404 or a 500?

runcom: yeah, probably 404

mmartinv (approved):
Looks good to me, thanks!

PR #675, opened 2024-09-30
fix(owner-tool): use the new API to export ovs

runcom:
This change allows the use of the new REST API to export either a single ov or multiple (tar-ed up).
Tested with the following:

```
$ ./target/debug/fdo-owner-tool export-manufacturer-vouchers --device-guid ba934a85-d55e-4930-0595-faeaa6090ddf http://localhost:8080
$ ./target/debug/fdo-owner-tool export-manufacturer-vouchers --device-guid ba934a85-d55e-4930-0595-faeaa6090ddf --path /home/amurdaca http://localhost:8080
$ ./target/debug/fdo-owner-tool export-manufacturer-vouchers http://localhost:8080 
```

### DIFF
--- a/test/fdo-postgres.sh
+++ b/test/fdo-postgres.sh
@@ -160,7 +160,9 @@ sudo podman run \
     --privileged \
     -v "$PWD"/export-ov:/export-ov:z \
     localhost/clients \
-    fdo-owner-tool export-manufacturer-vouchers postgres "postgresql://${POSTGRES_USERNAME}:${POSTGRES_PASSWORD}@${POSTGRES_IP}/${POSTGRES_DB}" /export-ov/ | grep "exported"
+    fdo-owner-tool export-manufacturer-vouchers "http://${FDO_MANUFACTURING_ADDRESS}:8080" --path /export-ov | grep "exported"
+sudo tar xvf "$PWD"/export-ov/export.tar -C "$PWD"/export-ov
+sudo rm -rf "$PWD"/export-ov/export.tar
 EXPORTED_FILE=$(ls -1 export-ov)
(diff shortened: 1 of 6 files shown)


runcom:
> I'd love to see some unit tests for this. We lacked unit tests in #673, too. We can do better! 💪

these functionalities are (to me) better covered with functional tests (which both this and #673 already have) - the remaining part is rust's warp wrapping which I'm not sure we need to cover (as we already run the tool, effectively excercising the paths here)

runcom:
ready for review now!

test/fdo-postgres.sh
```diff
@@ -160,7 +160,9 @@ sudo podman run \
     --privileged \
     -v "$PWD"/export-ov:/export-ov:z \
     localhost/clients \
-    fdo-owner-tool export-manufacturer-vouchers postgres "postgresql://${POSTGRES_USERNAME}:${POSTGRES_PASSWORD}@${POSTGRES_IP}/${POSTGRES_DB}" /export-ov/ | grep "exported"
+    fdo-owner-tool export-manufacturer-vouchers "http://${FDO_MANUFACTURING_ADDRESS}:8080" --path /export-ov | grep "exported"
```
mmartinv, line 163: Don't we return the corresponding exit code? I mean, why don't we check the exit code instead of the program output?

runcom: that's a good question - I just adapted to the new command w/o changing the logic

runcom: Will address this in a follow up!

runcom: See https://github.com/fdo-rs/fido-device-onboard-rs/issues/680

PR #676, opened 2024-10-02
fix(systemd-units): run before powering off the system

mmartinv:
Run before coreos-installer-poweroff.service and
coreos-installer-noreboot.service systemd units.

Otherwise the coreos-installer-poweroff.service and
the manufacturing-client.service are run in parallel
and the manufacturing client can be stopped before
finishing the manufacturing process.

Resolves: RHEL-38482


runcom (approved):
I think at some point we should (likely) move out the dracut module out of this repo - the reason being it's too tied to "other" things like coreos-installer

PR #677, opened 2024-10-02
fix(systemd-generator): write configuration to '/run'

mmartinv:
Write manufacturing-client-config to '/run' instead of
'/etc' as it seems to be a readonly filesystem.

Resolves: fedora-iot/iot-distro#64


runcom (approved):

nullr0ute:
The /run directory isn't persistent, is there a requirement to keep any of the bits written across reboots?

mmartinv:
> The /run directory isn't persistent, is there a requirement to keep any of the bits written across reboots?

No, the manufacturing client configuration is regenerated on each boot by the systemd generator based on the kernel cmdline

PR #687, opened 2024-10-10
fix: use centos-stream-9 target instead of epel-9

mmartinv:
Use the centos-stream-9 target as the epel-9 repos
don't seem to be up to date.

Remove the preparation steps enabling the copr repo
and installing the copr packages because it's not needed
and it's causing problems: Packit is supposed to enable the
copr repo and install all the packages contained in
that repo.

Use pipx instead of pip3 as suggested by the error
message to fix the postgres tests.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -30,7 +30,7 @@ jobs:
       job: copr_build
       trigger: pull_request
       targets:
-          epel-9: {}
+          centos-stream-9: {}
           fedora-latest-stable: {}
           fedora-latest: {}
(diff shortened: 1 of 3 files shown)


nullr0ute (approved):
LGTM

nullr0ute:
> Use the centos-stream-9 target as the epel-9 repos don't seem to be up to date.

For future reference EPEL uses the current release of RHEL as the base repo (so currently 9.4+updates) where as stream is the *next* release of RHEL so it's likely the version that users would expect, not out of date.

PR #692, opened 2024-11-20
fix(make-vendored-tarfile.sh): exclude idna tests with unicode points

runcom:
I imagine we could get rid of all the tests/ directories in every create but for now, this addresses rpmlint in centos/rhel. See https://artifacts.osci.redhat.com/testing-farm/935ce55a-9d6d-4d11-be0a-7a0babd942b7/work-rpminspect20if1pvh/rpminspect/execute/data/guest/default-0/rpminspect-1/data/viewer.html#unicode We usually waived this kind of errors in the past but this is a good time and easy patch to fix it.


pcdubs (approved):
LGTM

PR #694, opened 2024-11-25
chore(.packit.yaml): propose update when building

runcom:
very much like what we have in greenboot https://github.com/fedora-iot/greenboot/blob/main/.packit.yaml#L19C3-L38C29


mmartinv (approved):
lgtm

PR #695, opened 2024-11-26
chore(packit): add pre-sync to fix propose-downstream action

runcom:
fixed packit configuration to allow to build the vendor archive in propose downstream too - as it's bad to add conditionals around sources anyway https://docs.fedoraproject.org/en-US/packaging-guidelines/SourceURL/#_do_not_conditionalize_sources
So, if we want just one shared specfile, this is the way :mandalorian:

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -21,6 +21,15 @@ srpm_build_deps:
     - cargo
     - openssl-devel
 
+packages:
+    fido-device-onboard-fedora:
+        downstream_package_name: fido-device-onboard
+        upstream_package_name: fido-device-onboard
+    fido-device-onboard-centos:
+        downstream_package_name: fido-device-onboard
+        upstream_package_name: fido-device-onboard
+        pkg_tool: centpkg
+
 actions:
     pre-sync:
         - bash -c "./make-vendored-tarfile.sh ${PACKIT_PROJECT_VERSION}"
@@ -41,9 +50,6 @@ jobs:
           fedora-latest-stable: {}
           fedora-latest: {}
           fedora-rawhide: {}
-#          fedora-eln:
-#              additional_repos:
-#                  - https://kojipkgs.fedoraproject.org/repos/eln-build/latest/$basearch/
 
     - job: tests
       trigger: pull_request
@@ -55,16 +61,23 @@ jobs:
           fedora-latest-stable: {}
           fedora-latest: {}
           fedora-rawhide: {}
-#          fedora-eln: {}
 
     - job: sync_from_downstream
       trigger: commit
 
     - job: propose_downstream
       trigger: release
+      packages: [fido-device-onboard-fedora]
       dist_git_branches:
         - fedora-development
         - fedora-latest-stable
+
+    - job: propose_downstream
+      trigger: release
+      packages: [fido-device-onboard-centos]
+      dist_git_branches:
+        - c10s
+        - c9s
 
     - job: koji_build
       trigger: commit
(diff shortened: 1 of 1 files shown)


fido-device-onboard.spec
```diff
@@ -11,7 +11,9 @@ License:        BSD-3-Clause
 URL:            https://github.com/fdo-rs/fido-device-onboard-rs
 Source0:        %{url}/archive/v%{version}/%{name}-rs-%{version}.tar.gz
 # See make-vendored-tarfile.sh in upstream repo
+%if 0%{?el}
```
nullr0ute, line 14: Should be 0%{?rhel}

runcom: already tried and doesn't work either, will keep trying

nullr0ute: Yes, I also think the changes that landed yesterday were incorrect but they were merged before I could comment.

runcom: why is that? it's working besides a conditional for centos

runcom:
fixed - @mmartinv ptal

F41 MR https://src.fedoraproject.org/rpms/fido-device-onboard/pull-request/6
rawhide MR https://src.fedoraproject.org/rpms/fido-device-onboard/pull-request/5

runcom:
rebased to latest main - not sure what @nullr0ute meant with "incorrect changes" here https://github.com/fdo-rs/fido-device-onboard-rs/pull/695#discussion_r1858109816 but this is very good to go finally (and automation works :man_dancing: )

.packit.yaml
```diff
@@ -4,8 +4,12 @@
 specfile_path: fido-device-onboard.spec
 
 files_to_sync:
-    - fido-device-onboard.spec
-    - .packit.yaml
+    - src:
+      - patches/0001-Revert-chore-use-git-fork-for-aws-nitro-enclaves-cos.patch
+      - .packit.yaml
+      - fido-device-onboard.spec
+      - "fido-device-onboard-rs-*-vendor-patched.tar.xz"
```
mmartinv, line 11: Shouldn't this be `fido-device-onboard-rs-${PACKIT_PROJECT_VERSION}-vendor-patched.tar.xz`?

.packit.yaml
```diff
@@ -18,6 +22,9 @@ srpm_build_deps:
     - openssl-devel
 
 actions:
+    pre-sync:
+        - bash -c "./make-vendored-tarfile.sh ${PACKIT_PROJECT_VERSION}"
+        - bash -c "git restore Cargo.lock"
```
mmartinv, line 27: Why do you need to restore the Cargo.lock file?

mmartinv (commented):
Maybe we can take this opportunity to also include downstream updates for at least c10s now that it's supported? https://packit.dev/docs/configuration/upstream/propose_downstream#syncing-the-release-to-centos-stream

.packit.yaml
runcom: we don't have that variable exposed in this snippet - it's also documented https://packit.dev/docs/configuration/examples#custom-archive-creation-for-release-syncing (see also the thread linked in slack)

.packit.yaml
runcom: the vendor script updates that, I'm just bringing it back to "normal"

runcom:
> I honestly would get this done for 0.5.1 as it's been going out of scope from the original bug 😄 but we can tackle that immediately next to this :)

scratch that, I made it work :) pushing!

nullr0ute (approved):
Looks fine

PR #697, opened 2024-11-26
fix dependabot issues

mmartinv:
- **fix: rename commitlint configuration file**
- **fix: convert to ES module**
- **fix: increase maximum size of commit messages**


runcom (approved):

.github/commitlint.config.mjs
```diff
@@ -8,7 +8,7 @@ const validateBodyMaxLengthIgnoringDeps = async (parsedCommit) => {
       && body != null
       && body.includes('Updates the requirements on');
 
-  const bodyMaxLineLength = 100;
+  const bodyMaxLineLength = 1000;
```
nullr0ute, line 11: I think 1000 is a bit of overkill for a line length, it would make it pretty unreadable.

.github/commitlint.config.mjs
```diff
@@ -16,7 +16,7 @@ const validateBodyMaxLengthIgnoringDeps = async (parsedCommit) => {
   ]
```
nullr0ute, line 16: What is ES in the context of the commit message? Please make commits actually useful.

.github/commitlint.config.mjs
mmartinv: You are right and actually it's a mistake, my mind removed `Line` from `bodyMaxLinelength` so I took the full commit message and I counted the characters (more than 500) instead of counting the characters in the longest line (less than 200). We can go down to 200 I guess, I will open a new PR for that, sorry

.github/commitlint.config.mjs
mmartinv: >ES Modules was introduced in ECMAScript 6 in 2015. It was created to provide a standardized way to modularize JavaScript code. ES Modules is now widely supported by modern web browsers and Node.js.

I thought it would be easy to verify what "ES module" means related to a javascript file change with a quick google search.

PR #702, opened 2024-11-27
chore(.packit.yaml): allow committers to dist-git to trigger koji/bodhi

runcom:
When somebody runs `packit propose-downstream` from the CLI, the MRs to update the spec are made by the user itself and not `packit`. When we merge these MRs, koji and bodhi won't run automatically unless we set allowed_builders/allowed_pr_authors, which this patch does.


nullr0ute (approved):
LGTM

PR #704, opened 2024-11-27
Include the patched version of aws-nitro-enclaves-cose in the repo and packit enhancements

mmartinv:
The purpose of this patch is to incorporate the external an patched version of [aws-nitro-enclaves-cose](https://github.com/nullr0ute/aws-nitro-enclaves-cose/tree/0.4.0-branch) into the project's source code to avoid all the hacks of patching and unpatching the spec files, the Cargo.toml and the vendored sources back an forth when packaging as it has been shown to be fragile and is causing (and will cause) problems with dependantbot.

This patch also tries to centralize all the build logic in one file (the Makefile) and simplify the packit configuration.

The summary of changes woud be:

- chore: add aws-nitro-enclaves-cose 0.4.0 crate to the repo
- chore: fix aws-nitro-enclaves-cose errors
- chore: spec changes
- chore: use the new local copy of aws-nitro-enclaves-cose
- chore: add new Makefile targets
- chore: use the new Makefile targets in packit
- chore: remove unused patches and scripts
- chore: skip external crates from analisys
- chore: bump libcryptsetup-rs
- chore: ignore source code and vendor tarballs.
- chore: avoid clippy errors


fido-device-onboard.spec
```diff
@@ -10,9 +10,8 @@ License:        BSD-3-Clause
 
 URL:            https://github.com/fdo-rs/fido-device-onboard-rs
 Source0:        %{url}/archive/v%{version}/%{name}-rs-%{version}.tar.gz
-# See make-vendored-tarfile.sh in upstream repo
-Source1:        %{name}-rs-%{version}-vendor-patched.tar.xz
-Patch1:         0001-Revert-chore-use-git-fork-for-aws-nitro-enclaves-cos.patch
+# We need the vendor tar file for CentOS and RHEL builds
+#Source1:        %%{name}-rs-%%{version}-vendor.tar.xz
```
runcom, line 14: I guess we can just remove this right?

mmartinv: No, currently we cannot, the targets related to rpm/srpm creation and the `packit-create-vendor` target modify exactly that line in case of CentOS/RHEL builds to add the vendor tar file. I kept the commented line because it's easier to replace (and because it gives some information). If you want to get rid of it I can modify the Makefile to add the `Source1` line after `Source0` instead of replacing/uncommenting the existing line. It's quite simple actually.

runcom: yeah, uhm, I mean, the whole point is to _not_ conditionalize the sources in the spec file (similarly to the guidelines https://docs.fedoraproject.org/en-US/packaging-guidelines/SourceURL/#_do_not_conditionalize_sources) so that we can just have one spec file shared between RHEL/centos-stream/Fedora - it seems this is instead going in a different direction :thinking: maybe propose-downstream (pre-sync) should instead modify the spec to include Source1 for centos-stream and we have that conditional only there

mmartinv: Actually, it's possible and we can share the same SPEC file. In fact it will simplify things. The only downside is that we will need to upload the vendor tar file to Fedora ccache even when the vendor's source file is not needed.

runcom: yep, that downside if perfectly acceptable according to the guidelines (as I read them, correct me if I'm wrong)

Makefile
```diff
@@ -23,9 +25,45 @@ help:
 	@echo "The following targets are available:"
 	@echo
 	@echo "    help:               Print this usage information."
+	@echo "    source:             Generate source tar file in the current directory."
+	@echo "    vendor:             Generate vendor tar file in the current directory."
```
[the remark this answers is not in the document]
mmartinv: Shouldn't the `propose-downstream` also work for centos? (except the corresponding Jira ticket creation)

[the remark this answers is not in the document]
runcom: yeah, but you still have to upload the sources (this is how I read the page on the packit guide as it's not yet implemented to upload sources to the lookaside cache) - also, building isn't taken care of as far as I can tell 

maybe I'm wrong :)

nullr0ute:
This shouldn't really be one massive PR, there's a whole raft of different changes in here, I'm also not particularly happy about pulling an unrelated code base into here due to maintenance and CVEs. I feel it would likely be less work to just move to the latest version of the COSE crate so we don't have to do this malarkey and we'd not have the problem again moving forward.

mmartinv:
> This shouldn't really be one massive PR, there's a whole raft of different changes in here,

I can separate the crate changes from the packit changes if you want but that would mean to move one or two commits only.

> I'm also not particularly happy about pulling an unrelated code base into here due to maintenance and CVEs.

Only the RHEL builds will be using the included code base, the Fedora builds are supposed to keep using the upstream patched crate which is rpm packaged in Fedora, so the only difference is that you will be adding the patch in this repo instead of adding the patch in the repo you have in your personal fork for the RHEL builds.

>I feel it would likely be less work to just move to the latest version of the COSE crate so we don't have to do this malarkey and we'd not have the problem again moving forward.

It's not, I tried to update and we need changes in the code to implement additional traits for some particular types. Moreover, there are some crate additions in the the updated version so we would need to add some new RPMs to Fedora if we upgrade, not checked the full list of dependencies though. Taking into account that we agreed to migrate to `coset` crate I don't think the effort to update worth it.

mmartinv:
Closing as this is not going to be merged, moving useful parts to other PRs

Issue #684, opened 2024-10-08
Migrate from tempdir crate to something else, preferably in the stdlib


nullr0ute:
The last release of tempdir was 6+ years ago, it has a dependency on a specific version of remove_dir_all which now has security issues.

Migrate to something more up to date, there should be something in the stdlib which should suffice for this sort of functionality.

nullr0ute:
This might be enough: https://doc.rust-lang.org/std/env/fn.temp_dir.html

nullr0ute:
Fixed via #710

PR #710, opened 2025-01-06
chore: move from tempdir to tempfile

decathorpe:
The tempdir crate is obsolete - it has not been updated since 2017, and it has been officially marked as deprecated in [RUSTSEC-2018-0017](https://rustsec.org/advisories/RUSTSEC-2018-0017.html), with the `tempfile` crate being an obvious replacement.

It looks like only the "manufacturing-server" component still depends on `tempdir`, everything else already depends on the `tempfile` crate.  Moving this workspace member to `tempfile` too eliminates a lot of cruft from the lockfile, too.

From what I can tell, `tempfile::TempDir::with_prefix(prefix)` is a drop-in replacement for `tempdir::TempDir::new(prefix)`.

---

For context, I'm trying to drop the package for the `tempdir` crate from Fedora Linux, and fido-device-onboard is one of the few remaining packages that pull it in.


nullr0ute:
Thanks, we had been tracking this via #684 but no one had got to it yet.

nullr0ute:
Builds fine locally

nullr0ute (approved):
Looks fine, builds locally

PR #717, opened 2025-01-09
chore: Fix up the make vendor patch

nullr0ute:
Fix up the patch used in the make vendor script.


miabbott:
I believe there is an alternate proposal for the patched `aws-nitro-enclave-cose` crate here - https://github.com/fdo-rs/fido-device-onboard-rs/pull/704

Can we consider that idea before making additional changes to the patch we are carrying?

PR #721, opened 2025-01-23
chore: update CONTRIBUTING with additional deps

miabbott:
When trying to `cargo build` using Fedora 41, I needed to install the `sqlite` and `libpq` devel packages.

I also sync'ed the two lists of dependencies in the Fedora/toolbox examples for better consistency.


nullr0ute:
Minor clarification, it's not F-41 that changed this dep, but features added in FDO 0.5+

nullr0ute (approved):
LGTM

PR #743, opened 2025-02-12
packit: drop centos propose downstream and jobs

runcom:
as the title says

fixes https://github.com/fdo-rs/fido-device-onboard-rs/issues/742


nullr0ute (approved):
LGTM, never did know why we did that

PR #745, opened 2025-02-17
Makefile tarball targets

mmartinv:
* Add two new `source-tarball` and `vendor-tarball` targets to generate the corresponding tarballs.
* Use the `vendor` target just to generate the filtered vendor directory and configure the cargo builds accordingly.
* Chain all the bash commands in the `vendor` target so all the bash variables contain the expected values.
* Adjust packit configuration to use the new `vendor-tarball`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,9 @@ help:
 	@echo "The following targets are available:"
 	@echo
 	@echo "    help:               Print this usage information."
-	@echo "    source:             Generate source tar file in the current directory."
-	@echo "    vendor:             Generate vendor tar file in the current directory."
+	@echo "    source-tarball:     Generate source tar file in the current directory."
+	@echo "    vendor-tarball:     Generate vendor tar file in the current directory."
+	@echo "    vendor:             Vendor dependencies and configure cargo build accordingly."
 	@echo "    rpm:                Generate RPM."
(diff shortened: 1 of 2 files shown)


runcom (approved):

runcom:
there's a typo :smile:

Makefile
```diff
@@ -42,34 +43,39 @@ SOURCE_TARBALL=fido-device-onboard-rs-$(VERSION).tar.gz
...
+	for platform in $(PLATFORMS); do  \
+		args+="--platform $${platform} "; \
+	done; \
+	# https://issues.redhat.com/browse/RHEL-65521 \
+	args+="--exclude-crate-path idna#tests "; \
+	rm -rf vendor; \
+	# Use the official crate version \
+	patch -p1 < patches/0001-Revert-chore-use-git-fork-for-aws-nitro-enclaves-cos.patch; \
+	mkdir -p .cargo; \
+	cargo vendor-filterer $${args} > ./.cargo/config.toml; \
```
runcom, line 74: what do we need config.toml for :thinking:

mmartinv: To use the vendored files as the source for cargo builds. If you don't specify that in `.cargo/config.toml` then the vendored files are not used.

runcom (approved):

PR #748, opened 2025-02-27
chore: .packit.yaml: do not put release changelog in spec changelog

runcom:
It's just too verbose...


nullr0ute:
> What about writing a custom changelog entry like described in [Update dist-git content](https://packit.dev/docs/fedora-releases-guide#update-dist-git-content)?

Yes, I think that looks good.

PR #762, opened 2025-04-26
fix: replace default.target with initrd.target

### LINKED ISSUE #761: Use of `default.target` breaks initramfs build on Fedora 42
The use of `default.target` in `module-setup.sh` (https://github.com/fdo-rs/fido-device-onboard-rs/blob/main/dracut/52fdo/module-setup.sh#L22) breaks initramfs build on Fedora 42, which in turn causes Fedora IoT Simplified provisioner to fail to build, see https://github.com/fedora-iot/iot-distro/issues/89.

lbrabec:
`default.target` is not present in c10s, RHEL10, and Fedora 42 initrd so it breaks the initramfs build.

Fixes #761

### DIFF
--- a/dracut/52fdo/module-setup.sh
+++ b/dracut/52fdo/module-setup.sh
@@ -19,5 +19,5 @@ install() {
         "/usr/libexec/manufacturing-client-service"
 
     install_and_enable_unit "manufacturing-client.service" \
-        "default.target"
+        "initrd.target"
 }


nullr0ute:
Generally looks fine, when was initrd.target introduced?

lbrabec:
Introduced in systemd: https://github.com/systemd/systemd/pull/14177
Change in dracut that removed `default.target`: https://github.com/dracut-ng/dracut-ng/pull/627

PR #765, opened 2025-05-16
ci: add Image Mode test automation

mcattamoredhat:
Enable CI test in RHEL-10 Image Mode.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -29,6 +29,10 @@ packages:
     downstream_package_name: fido-device-onboard
     upstream_package_name: fido-device-onboard
     pkg_tool: centpkg
+  fido-device-onboard-rhel:
+    downstream_package_name: fido-device-onboard
+    upstream_package_name: fido-device-onboard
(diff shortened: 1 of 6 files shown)


.packit.yaml
```diff
@@ -101,3 +101,14 @@ jobs:
...
+
+  # RHEL-10 jobs
+
+  - job: tests
+    trigger: pull_request
+    identifier: onboarding-image-mode
+    fmf_path: test/fmf
+    tmt_plan: plans/image-mode-onboarding
+    use_internal_tf: true
+    packages: [fido-device-onboard-centos]
```
mmartinv, line 113: Maybe I am wrong but if you are going to build the package in the spawned VM you don't need to specify `packages`, just use `skip_build: true` to run the tests without a copr build.

mcattamoredhat: I'm skipping package building in the spawned VM, to make use of the copr build. But the issue I'm having is that rhel-10 job is not being triggered.

Issue #763, opened 2025-04-30
Fix clippy warning


nullr0ute:
Seeing this on CI runs.

```
    Checking retry v1.3.1
error: manual implementation of `ok`
   --> util/src/servers/mod.rs:122:20
    |
122 |           Some(v) => match v.into_string() {
    |  ____________________^
123 | |             Ok(s) => Some(s),
124 | |             Err(_) => None,
125 | |         },
    | |_________^ help: replace with: `v.into_string().ok()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_ok_err
    = note: `-D clippy::manual-ok-err` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::manual_ok_err)]`

   Compiling fdo-data v0.5.4 (/__w/fido-device-onboard-rs/fido-device-onboard-rs/libfdo-data)
error: could not compile `fdo-util` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

pcdubs:
This is fixed.

PR #771, opened 2025-05-27
chore: bump for 0.5.5 release

pcdubs:
Prepare for the 0.5.5 release.


nullr0ute:
The clippy warning is tracked by #763 and isn't fatal but code needs to be updated for the latest rust style.

The build and test failure is because the fdo components weren't bumped in the Cargo.lock file. This would have been picked up with a local `cargo build --release` and checking the diff once the versions were bumped.

There also needs to be a changelog entry added to the rpm spec file.

PR #775, opened 2025-05-28
chore: fix the container deps

nullr0ute:
The aio (All In One) container is missing the DB client libraries and the deps for the FDO client to run, while at it drop the -devel packages for the services as they only need the actual libraries to run.


pcdubs (approved):
LGTM

PR #779, opened 2025-07-07
ci: fix clippy check spelling and labeler failures

mcattamoredhat:
In order to fix Clippy issue, the strict warning treatment has been removed.

Some exceptions have been added to `.github/spellcheck-ignore` file.

`.github/labeler.yml` file has been updated to align with changes of Pull Request labeler v5 actions/labeler@v5 (add missing quotes and use the changed-files structure)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,3 +1,4 @@
-possible stability impact:
-- libfdo-data/fdo_data.h
-- libfdo-data/libfdo-data-go.doc
+"possible stability impact":
+  - changed-files:
+    - any-glob-to-any-file: "libfdo-data/fdo_data.h"
+    - any-glob-to-any-file: "libfdo-data/libfdo-data-go.doc"
--- a/.github/spellcheck-ignore
+++ b/.github/spellcheck-ignore
@@ -6,3 +6,6 @@ childs
 ot
 marshalling
 te
+blacklist
+msdos
+ro
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
         with:
           toolchain: stable
           components: clippy
-      - run: cargo clippy -- -D warnings -D clippy::panic -D clippy::todo
+      - run: cargo clippy -- -D clippy::panic -D clippy::todo
 
   build_and_test:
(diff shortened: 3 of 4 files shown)


miabbott:
@mcattamoredhat This is probably annoying, but could you split the changes into separate commits for each CI check you have fixed?

.github/workflows/ci.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         with:
           toolchain: stable
           components: clippy
-      - run: cargo clippy -- -D warnings -D clippy::panic -D clippy::todo
+      - run: cargo clippy -- -D clippy::panic -D clippy::todo
```
7flying, line 82: Since we don't have Rust veterans in the team I would encourage to maintain clippy failing on warnings, we need more learning opportunities

mcattamoredhat:
> @mcattamoredhat This is probably annoying, but could you split the changes into separate commits for each CI check you have fixed?

Sure, no problem @miabbott!

.github/workflows/ci.yml
mcattamoredhat: Ok @7flying I understand. These warnings are related to the older format string syntax usage in rs files

I.e. https://github.com/fdo-rs/fido-device-onboard-rs/blob/main/util/src/servers/mod.rs#L73 
`format!("{}.yml", guid)`
Instead of 
`format!("{guid}.yml")`

There is more than one file using this syntax. Thus, should we change all of them (despite those changes wouldn't be related to Clippy command) in a separated PR?

.github/spellcheck-ignore
```diff
@@ -6,3 +6,6 @@ childs
 ot
 marshalling
 te
+blacklist
```
nullr0ute, line 9: this should be a separate commit.

.github/labeler.yml
```diff
@@ -1,3 +1,4 @@
-possible stability impact:
-- libfdo-data/fdo_data.h
-- libfdo-data/libfdo-data-go.doc
+"possible stability impact":
```
nullr0ute, line 1: Again a separate commit, generally each fix should be independent, make it easier for bisect and review.

PR #784, opened 2025-07-08
ci: add ci fixes

mcattamoredhat:
This PR gathers the changes of the following PRs: https://github.com/fdo-rs/fido-device-onboard-rs/pull/780, https://github.com/fdo-rs/fido-device-onboard-rs/pull/781, https://github.com/fdo-rs/fido-device-onboard-rs/pull/783

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,3 +1,5 @@
-possible stability impact:
-- libfdo-data/fdo_data.h
-- libfdo-data/libfdo-data-go.doc
+"possible stability impact":
+  - changed-files:
+    - any-glob-to-any-file:
+      - "libfdo-data/fdo_data.h"
+      - "libfdo-data/libfdo-data-go.doc"
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -10,7 +10,12 @@ jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
+    - uses: actions/checkout@v4
+      with:
+        repository: "fdo-rs/fido-device-onboard-rs"
+        ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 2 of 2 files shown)


nullr0ute (dismissed):
Overall looks fine to me, I would like someone with a better rust knowledge than me to review the clippy fixes too

nullr0ute:
and looking at the labeler CI failure it appears to still not like that

.github/labeler.yml
```diff
@@ -1,3 +1,4 @@
-possible stability impact:
-- libfdo-data/fdo_data.h
-- libfdo-data/libfdo-data-go.doc
+"possible stability impact":
+  - changed-files:
+    - any-glob-to-any-file: "libfdo-data/fdo_data.h"
+    - any-glob-to-any-file: "libfdo-data/libfdo-data-go.doc"
```
miabbott, line 4: This should be a YAML array now; [see the new syntax](https://github.com/actions/labeler?tab=readme-ov-file#match-object)

```suggestion
    - any-glob-to-any-file: 
        - "libfdo-data/fdo_data.h"
        - "libfdo-data/libfdo-data-go.doc"
```

7flying:
Rust part LGTM.

The postgres_test part of the CI needs to be re-launched, it should pass

nullr0ute:
So I think this is still failing because it's using the main branch for the config file AFAICT from a couple of tickets I had found and the [Changes page](https://deepwiki.com/actions/labeler/3-configuration). So I think this should be fine, we can always do a follow up.

nullr0ute:
Some further details on the [pull_request_target event](https://github.com/actions/labeler#notes-regarding-pull_request_target-event) as it runs on the base branch FYI.

nullr0ute:
That worked on re-running CI on another merge request :+1:

PR #803, opened 2026-03-15
Update nix dependency from 0.26 to 0.31

musicinmybrain:
https://github.com/nix-rust/nix/blob/v0.31.2/CHANGELOG.md

No source-code changes appear to be required, although I did have to explicitly enable some features. I tested this by running `cargo test` in `client-linuxapp/` and in `admin-tool/`.

The motivation is that, in Fedora, we’re trying to reduce dependencies on our `rust-nix0.26` compat package, for reasons described in the [downstream issue](https://forge.fedoraproject.org/rust/backlog/issues/7), and it’s our policy to [offer such changes upstream](https://docs.fedoraproject.org/en-US/package-maintainers/Staying_Close_to_Upstream_Projects).


nullr0ute (approved):
LGTM

PR #809, opened 2026-07-29
fix(.packit.yaml): use explicit distros mapping for CentOS Stream tests

mmartinv:
Use dict format with explicit distros in the tests job to prevent
packit from incorrectly mapping centos-stream targets to epel chroots.

Workaround for https://github.com/packit/packit/issues/2700


nullr0ute (approved):
LGTM, not really up on centos though

